PR #23, opened 2015-03-31
Add a driver for the PCA9658 16-channel, 12-bit PWM controller.

cmnrd:
A description that is a bit more detailed can be found in the comments of the source files. This implementation is rather basic as it does not implement all features. However, it should be sufficient for most use cases. The code is tested and I'm using it for my ambilight project.

### DIFF
--- a/src/xpcc/driver/pwm/pca9685.hpp
+++ b/src/xpcc/driver/pwm/pca9685.hpp
@@ -1,0 +1,239 @@
+// coding: utf-8
+// ----------------------------------------------------------------------------
+ /*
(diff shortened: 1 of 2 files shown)


salkinium:
Your code looks good, do you have a simple program that you can add to `examples/` with the platform of your choice?

This way Travis CI will be able to test the class and we have one more example ;-)

cmnrd:
Yes I will provide something, I have to cleanup and simplify my code anyway. So it shouldn't be too hard to get a small example on the way ;)

Issue #29, opened 2015-08-17
Enable C++14 on all platforms


ekiwi:
C++14 has been out for more than half a year now, and provides some handy features like
[Relaxed constexpr restrictions](https://en.wikipedia.org/wiki/C%2B%2B14#Relaxed_constexpr_restrictions).

GCC 5 has (at least for our purposes) full [C++14 support](https://gcc.gnu.org/projects/cxx1y.html). Unfortunately it is not easily available for all available platforms right now. I suggest we use this issue to track on what platforms GCC 5 is missing.

**Hosted** (`gcc`)
- Fedora 22 and up comes with at least `gcc 5.1.1`
- Ubuntu 15.04?
- Homebrew?
- Travic-CI?

**ARM** (`arm-none-eabi-gcc`)
- The prebuilt `arm-none-eabi` toolchain that we use on all platforms will release [arm-none-eabi-gcc 5.0](https://launchpad.net/gcc-arm-embedded/+milestone/5.0-2015-q4-major) in December 2015

**AVR** (`avr-gcc`)
- Fedora 23 (which hasn't been released) will feature [avr-gcc 5.1.0](https://apps.fedoraproject.org/packages/avr-gcc)
- Ubuntu?
- Homebrew?
- What about the atmel packages that we use on travis-ci?

dergraaf:
Debian 8 has gcc 4.9.2, avr-gcc 4.8.1 and arm-none-eabi-gcc 4.8.4.

salkinium:
For avr-gcc on travis-ci, we [download the latest toolchain from Atmel](https://github.com/roboterclubaachen/xpcc/blob/develop/.travis.yml#L17), since it is not available on the old Ubuntu that travis-ci uses.

ekiwi:
> For avr-gcc on travis-ci, we download the latest toolchain from Atmel, since it is not available on the old Ubuntu that travis-ci uses.

I just added some commands to our `.travis.yml` to display the gcc version. [Apparently](https://travis-ci.org/roboterclubaachen/xpcc/jobs/76096308#L515), the Atmel build that we use is a `gcc 4.8.1`.

dergraaf:
On a similar note: What version of gcc is required to build the current development branch? The arm-none-eabi-gcc 4.8.2 I am using seems to be to old.

salkinium:
[...]

PR #40, opened 2015-11-15
Implement sized deallocation operators for c++14 compatibility

cmnrd:
I used C++11 on the AVR before, but of lately my compiler complains about undefined reference to `operator delete(void*, unsigned int)'. This operator is needed for the sized deallocation feature of C++11. This fix implements the new delete operators by simply dropping the size parameter. However, maybe the additional parameter could be used for more efficient implementation of freeMemory.

### DIFF
--- a/src/xpcc/architecture/platform/driver/core/avr/newdelete.cpp
+++ b/src/xpcc/architecture/platform/driver/core/avr/newdelete.cpp
@@ -28,7 +28,19 @@ operator delete(void* ptr)
 }
 
 void
+operator delete(void* ptr, size_t size __attribute__((unused)))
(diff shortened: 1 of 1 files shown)


ekiwi:
Looks good. Thank you.

cmnrd:
Just noticed that it's a C++14 issue, not 11.

salkinium:
Looks good to me.
Paging @dergraaf since he implemented the memory management on the AVR.

I can't wait to have C++14 support for xpcc.
I think we are just waiting for arm-none-eabi-gcc to publish their precompiled version (see #29) in Dec. 2015.

PR #39, opened 2015-11-07
Improve SystemClock implementation for STM32

salkinium:
In this feature I rewrote the SystemClock templates to take input from a clock tree description in the device file and generate the appropriate template classes.
### Summary
- [x] Rewrite and generalize `ClockControl` class
- [x] Implement generic Source class
- [x] Implement generic Sink class
- [x] Implement generic Simple PLL class
- [ ] Implement generic Complex PLL class
- [ ] Move STM32 clock tree parsing code from [here](https://github.com/salkinium/save-the-clocktrees) into device file generator
- [ ] Implement generic STM32 clock tree parsing to auto-generate the clock trees of all STM32 devices
### Details

There currently are three classes that can be generated:
- clock sources,
- clock sinks, and
- simple PLLs (`Output = Input * Multiplier / Divisor`).

All classes conform to the following naming scheme of static member variables and methods:

``` cpp
static const TypeId::TypeIdName Id;
static constexpr ClockName Name;
static constexpr uint32_t InputFrequency;
static constexpr uint32_t OutputFrequency;
static StartupError enable(const uint32_t waitCycles = 1500);
```

The `ClockName` is a substitute for a type system, which enforces the allowed clock inputs for any tree node at compile time using static asserts.
This makes it possible to check this without template specializations, and therefore make it possible to use the same class regardless of tree length without duplicate code.

The `ClockTree` class has been renamed to `ClockControl` and the following naming scheme is introduced:
- `static bool enable{{SourceName}}Clock(uint32_t waitCycles = 1500);` for clock sources
- `static bool enable{{SinkName}}Clock({{SinkName}}ClockSource src, uint32_t waitCycles = 1500);` for clock sinks
- `static bool enablePll(PllSource source, ..., uint32_t waitCycles = 1500);`, different for every PLL

Typical usage of the SystemClock:

``` cpp
using systemClock = SystemClock< Pll< ExternalCrystal<MHz8>, MHz24 > >;
StartupError error = systemClock::enable();
```

The following examples are from the STM32F100 clock tree, but similar properties can be found on other STM32 devices.
#### Sources

``` xml
<!-- High speed clocks -->
<source name="InternalClock" type="clock" location="internal" speed="high" fixed="8000000"/>
<source name="ExternalClock" type="clock" location="external" speed="high" min="1000000" max="24000000"/>
<source name="ExternalCrystal" type="crystal" location="external" speed="high" min="4000000" max="24000000"/>
<!-- Low speed clocks -->
[...]

PR #41, opened 2015-11-20
Update STM32F4 header and device files

salkinium:
- Adds required header files from CubeF4v2.4
- Updates device file reader, merger and auxilary information
- Adds device files for all STM32F4 devices.


ekiwi:
Do we really want to add device files for devices that we do not own and thus cannot verify the support?
Currently I look at the available device files to determine which controllers should work with `xpcc`, but if we add all F4 devices, that might not be true anymore.

dergraaf:
Perhaps a good solution is to provide a table with an overview about what has been verified in Hardware and when.
Only adding devices we own is not usefull longterms, but adding too many devices with no one to support them is worse.

salkinium:
We kinda have a table of devices with first-class support in the [Readme](https://github.com/roboterclubaachen/xpcc#hardware-support).
However, I will add an explicit file with more details about this.

It's a little late for not adding a massive amount of device file, as I've already added _all_ AVR AT90, ATtiny, ATmega and Xmega device files…
(We don't even have Xmega drivers for anything).

The reasoning for add device files for devices we haven't tested it yet was, that generating the device files is not trivial, so someone new who might have a device we do not own, has practically no chance of even starting off.
So even if the support isn't good, there is a larger chance for someone unfamiliar with xpcc to start contributing to it.
Also, I just bought the STM32F469-Disco and I expect most peripherals to be similar if not identical.

I want to extract the device file generator out of xpcc into it's own repo and base it on [NetworkX](http://networkx.github.io) and therefore improve the merging algorithm.
The current algorithm only really works on flat, attributed lists, which doesn't work anymore on something very connected like clock trees.

Let's not kid ourselves here: We work with CAP: Computer Assisted Porting.
We could, with the right tools, provide relatively stable support for all STM32F\* devices, with reasonably little work.
Most ports are island solutions where one person takes exactly one variation of one device and copy and pastes a bunch of code.
I have not seen anyone else take such an algorithmically assisted and complete approach to porting, because it's not trivial. 
This is really exciting!

salkinium:
Well, the STM32F469 Disco just works out the box. So I'm fairly confident the others also work.
Except for the SystemClock, of course, that is being addressed by #39.

PR #42, opened 2015-11-23
Driver: LSM303: Remove stray reference to LIS.

strongly-typed:
Probably from copy & paste.

### DIFF
--- a/src/xpcc/driver/inertial/lsm303a.hpp
+++ b/src/xpcc/driver/inertial/lsm303a.hpp
@@ -13,7 +13,6 @@
 #include <xpcc/architecture/interface/register.hpp>
 #include <xpcc/processing/resumable.hpp>
 #include <xpcc/math/utils/endianness.hpp>
-#include "lis3_transport.hpp"
 
 namespace xpcc
 {


salkinium:
Are you really sure that's a stray reference?

The LSM303A shares the same protocol over I2C as all other LIS3\* based sensors.
So the sensor driver quite literally inherits from [`Lis3TransportI2c`](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/driver/inertial/lsm303a.hpp#L390) :wink: 

I think the CI only passes, since the [gyroscope L3GD20 header](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/driver/inertial/l3gd20.hpp#L16) includes the `lis3_transport.hpp` [just before](https://github.com/roboterclubaachen/xpcc/blob/develop/examples/stm32f3_discovery/stm32f3_discovery.hpp#L20).

I hope you don't mind if I let Dr. Cox answer this one :stuck_out_tongue_winking_eye:
[NOOOOOOOOOOO](https://www.youtube.com/watch?v=3zNjQecyjE8)

salkinium:
Bonus: This shall forever be remembered as #42 of xpcc.

PR #43, opened 2015-11-25
Add STM32F469I-DISCO example

salkinium:
Simply blinks and prints a loop counter via the STLink v2.1 serial port.

You need to compile openocd from source to use this, since v0.9.0 does not contain F469 support yet!
OSX: `$ brew install --HEAD openocd`

I think it makes sense to declare the `xpcc::logger::*` streams directly in the board support header, right?
The ST-Link has an integrated serial port and the pins for it are already dedicated, so not using it for the serial stream wouldn't gain you anything.


ekiwi:
> I think it makes sense to declare the xpcc::logger::\* streams directly in the board support header, right?

In general the rational was to show the lowest memory usage possible for the examples as not to discourage people from adopting `xpcc`.

Using the `xpcc::logger` subsystem causes the memory allocator to be linked and thus the memory usages goes up by quite a bit.
However, this is probably ok for STM32F4 platforms.

ekiwi:
> I think it makes sense to declare the xpcc::logger::\* streams directly in the board support header, right?

Oh and on a second thought: you shouldn't put the streams into the board support header, as this would lead to linker errors if you include the file in more than one compilation unit.

salkinium:
> In general the rational was to show the lowest memory usage possible for the examples as not to discourage people from adopting xpcc.
> Using the xpcc::logger subsystem causes the memory allocator to be linked and thus the memory usages goes up by quite a bit.
> However, this is probably ok for STM32F4 platforms.

Well. This board has 320kB SRAM + 64kB CCM and a whopping 16MB of external RAM.
I think we would have great difficulty filling up even just the internal RAM, so for this board it's probably ok.

> Oh and on a second thought: you shouldn't put the streams into the board support header, as this would lead to linker errors if you include the file in more than one compilation unit.

Yeah, exactly. It would need to be placed into a `.cpp` file, but how do I get that compiled?

``` python
files = env.FindFiles('.') + '../stm32f469_discovery.cpp'
```

That doesn't work though and it is quite ugly.

ekiwi:
> files = env.FindFiles('.') + '../stm32f469_discovery.cpp'
> 
> That doesn't work though and it is quite ugly.

Still ugly, but does:

```
files = env.FindFiles('.') + ['../stm32f469_discovery.cpp']
```

work?

salkinium:
No. [`FindFiles` returns a Scanner object](https://github.com/roboterclubaachen/xpcc/blob/develop/scons/site_tools/configfile.py#L282-L287), which has a [method called `append()`](https://github.com/roboterclubaachen/xpcc/blob/develop/scons/site_tools/configfile.py#L198-L205), which seems to do the exact thing we need:

``` python
files = env.FindFiles('.').append('../stm32f469_discovery.cpp')
```

However, it [calls `FileList.append`](https://github.com/roboterclubaachen/xpcc/blob/develop/scons/site_tools/configfile.py#L86-L90), which calls `FileList.extend`, which calls `FileList.append`, which calls `FileList.extend`, which… you get the picture.
Changing [this](https://github.com/roboterclubaachen/xpcc/blob/develop/scons/site_tools/configfile.py#L101) to `self.__append(item)` then results in [an error here](https://github.com/roboterclubaachen/xpcc/blob/develop/scons/site_tools/configfile.py#L94): 

```
File "/Users/niklas/development/xpcc_cube_update/scons/site_tools/configfile.py", line 94:
    item = SCons.Node.FS.default_fs.File(str(item))
  File "/usr/local/Cellar/scons/2.3.5/libexec/scons-local/SCons/Node/FS.py", line 1340:
    return self._lookup(name, directory, File, create)
  File "/usr/local/Cellar/scons/2.3.5/libexec/scons-local/SCons/Node/FS.py", line 1319:
    return root._lookup_abs(p, fsclass, create)
  File "/usr/local/Cellar/scons/2.3.5/libexec/scons-local/SCons/Node/FS.py", line 2252:
    result.must_be_same(klass)
  File "/usr/local/Cellar/scons/2.3.5/libexec/scons-local/SCons/Node/FS.py", line 627:
    (self.__class__.__name__, self.path, klass.__name__))
```

This is a wider question of how we want to provide Board Support Packages in the long run.
I think this can only cleanly be done using the library builder.

ekiwi:
Sorry. You can try:

``` .py
sources = env.FindFiles('.'). sources + ['../stm32f469_discovery.cpp']

program = env.Program(target = env['XPCC_CONFIG']['general']['name'], source = sources)
```

salkinium:
Yep, that works, but it feels very wrong.

PR #44, opened 2015-12-01
Feature/rca develop

strongly-typed:


salkinium:
This the is the correction of an unfortunate diversion between our internal fork of our public xpcc repository.

ekiwi:
We probably should have squashed the three commits into one.
To me there is no apparent reason, why these should be three separate commits.

salkinium:
Tell that to the author.
I cannot modify the RCA remote history without force pushing to the internal remote.

PR #46, opened 2015-12-08
Add STM32F7-DISCO example

salkinium:
Analog to #43 this simply blinks and prints a loop counter via the STLink v2.1 serial port.

You need to compile openocd from source to use this, since v0.9.0 does not contain F7 support yet!
OSX: `$ brew install --HEAD openocd`

Be aware that because no generic SystemClock support is yet available, the SystemClock is mocked up and the entire board operates on 16MHz.
Both data and instruction cache and flash prefetch are turned off, and the delay loop runs three times as fast as it should (the Cortex-M7 is black magic).


ekiwi:
> Be aware that because no generic SystemClock support is yet available, the SystemClock is mocked up and the entire board operates on 16MHz.

We really need to stop doing this....

What if I told you that we cannot merge this request, without a proper clock driver?
Would that help the development of your new feature to generate clock code? ;-)

salkinium:
> What if I told you that we cannot merge this request, without a proper clock driver?

That's why it has the "in progress" tag. I don't plan to merge this right now, this PR is to create awareness and start a discussion.
Like we just did ;-)

On the other hand: We can create terrible SystemClock mockups.
And it works, because compile-time duck-typing !1!!

> Would that help the development of your new feature to generate clock code? ;-)

Grrr. Yes, the feature is indeed quite desperately needed.

I actually first wanted to finish and merge the feature/safe_stack, but I had more unstaged changes in the repo, so here we are.

ekiwi:
> Grrr. Yes, the feature is indeed quite desperately needed.

At the robotics club, we are currently designing a new PCB using the STM32F072 controller.
Thus I'll have to mess with its clock tree soon.
I do not remember what it was, but there was one clock source that just wouldn't fit into the current system.

salkinium:
One caveat remains: I cannot get the DWT cycle counter to start counting, the counter simply remains at zero. So the delay functions will be busy waiting instead of polling.

salkinium:
The SystemClock is mocked up in this PR, however, considering the complexity of a automated SystemClock implementation, this is the best solution currently.

PR #47, opened 2015-12-12
Travis CI: Enable container-based virtualization environment

salkinium:
Our current CI is running in the (legacy) standard virtualization environment, which does not have as [much computational resources as the container-based one](https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments).
Furthermore caching of directories is not available for public repositories, which means that for every instance in our unittests at least the AVR toolchain has to be downloaded again from Atmels servers, which adds time.

This feature enables the container-based environment and the use of the cache for toolchains and boost.
Since `sudo` is not available and [only whitelisted packages are allowed](https://github.com/travis-ci/apt-package-whitelist), we have to download both the AVR and Cortex-M toolchains, as well as build the latest boost package from source.
The results are cached, so that this only has to happen once.

Total running time is reduced from ~90min to about ~52min, with the elapsed time reducing from ~30min to about ~14min.
So the CI runs about twice as fast now, which is a great improvement us and also for Travis CI.

### DIFF
--- a/scons/site_tools/hosted.py
+++ b/scons/site_tools/hosted.py
@@ -101,5 +101,12 @@ def generate(env, **kw):
 		"-Woverloaded-virtual",
 	]
 
+	if 'CXXFLAGS' in os.environ:
+		flags = "".join(os.environ['CXXFLAGS'])
+		env['CXXFLAGS'].append(flags)
+	if 'LDFLAGS' in os.environ:
+		flags = "".join(os.environ['LDFLAGS'])
+		env['LINKFLAGS'] = flags
+
 def exists(env):
 	return env.Detect('g++')
--- a/.travis.dep.sh
+++ b/.travis.dep.sh
@@ -1,0 +1,30 @@
+cd $HOME/cache
+if [ ! -d "$HOME/cache/avr8-gnu-toolchain-linux_x86" ]; then
+	echo "Downloading AVR toolchain..."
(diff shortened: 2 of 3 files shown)


scons/site_tools/hosted.py
```diff
@@ -101,5 +101,12 @@ def generate(env, **kw):
 		"-Woverloaded-virtual",
 	]
 
+	if 'CXXFLAGS' in os.environ:
+		flags = "".join(os.environ['CXXFLAGS'])
+		env['CXXFLAGS'].append(flags)
+	if 'LDFLAGS' in os.environ:
+		flags = "".join(os.environ['LDFLAGS'])
+		env['LINKFLAGS'] = flags
+
```
salkinium, line 110: @ekiwi: Is there a better way to do this?

ekiwi, line 110: you could try `env.Append(ENV = {'PATH' : os.environ['PATH']})` from [line 56 of the same file](https://github.com/salkinium/xpcc/blob/fa6a986aea37b879e9101a1b36f12888970d2dcd/scons/site_tools/hosted.py#L56).

But in general it is what we are already doing, just with the other environment variables.

salkinium, line 110: Nah, I'm happy that it works now, so I'm not going to experiment further.
Build systems are not my thing ;-P

PR #49, opened 2016-01-07
[io] add overloaded printf functions that takes a va_list

ekiwi:
This is very similar to std::vprinf:
http://en.cppreference.com/w/cpp/io/c/vfprintf

It makes it possible for a varargs function to
relay its arguments to printf.


ekiwi:
I found it convenient to just overload `printf`, but we could also name it `vprintf` if we want to stick closer to the standard library.

salkinium:
Lets keep this closer to what people know, they are already confused enough with the rest ;-)
So lets rename it to `vprintf`.
Other than that :+1: from me.

PR #51, opened 2016-01-08
[math] BitOperations: rbit operation is not available on cortex-m0

### LINKED ISSUE #50: BitOperation: bitReverse does not work on cortex-m0
Apparently the [inline assembly](https://github.com/roboterclubaachen/xpcc/blob/3aaf58192c005ddea2478dc51c7faa2071f7746d/src/xpcc/math/utils/bit_operation.hpp#L171) used in the `bitReverse` function does not work on cortex-m0.

I get the following error from the compiler:

```
/tmp/ccykPtaf.s: Assembler messages:
/tmp/ccykPtaf.s:194: Error: selected processor does not support Thumb mode `rbit r1,r1'
/tmp/ccykPtaf.s:261: Error: selected processor does not support Thumb mode `rbit r3,r3'
```

ekiwi:
fixes #50

### DIFF
--- a/src/xpcc/math/utils/bit_operation.hpp
+++ b/src/xpcc/math/utils/bit_operation.hpp
@@ -166,7 +166,7 @@ namespace xpcc
 	inline uint8_t
 	bitReverse(uint8_t n)
 	{
-#ifdef XPCC__CPU_ARM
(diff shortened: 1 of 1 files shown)


salkinium:
:+1:

PR #53, opened 2016-01-08
[driver] STM32: Uart: name all files uart to prevent confusion

ekiwi:
Naming all files generated by the driver the same
cleans up the code and makes it easier to cope with
STM32F0 where all UART IPs are called USART.

As these files are not supposed to be included by the user directly
and all class names are kept the same, this should not
impact any existing code.

### DIFF
--- a/src/xpcc/architecture/platform/driver/core/cortex/hard_fault_handler.cpp.in
+++ b/src/xpcc/architecture/platform/driver/core/cortex/hard_fault_handler.cpp.in
@@ -54,8 +54,7 @@
 		%%	set uart = "Uart"
 	%% endif
 
-#include "../../uart/stm32/{{ uart | lower }}_{{ id }}.hpp"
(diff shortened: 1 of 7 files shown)


salkinium:
Don't forget to rename the `#include` in the [hard fault handler here](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/architecture/platform/driver/core/cortex/hard_fault_handler.cpp.in#L49-L74).
(Travis CI caught that one).

ekiwi:
@salkinium any idea, why travis seems to sometimes miss the compiler?

salkinium:
We moved to the containerized environment in #47, which means no more sudo.
So we install everything from source (even compile boost from source), which obviously only makes sense when using the cache.

I think this is an internal Travis problem with their cache.
A quickfix is to restart the errored job, and it will probably work (like I did above.)

salkinium:
:+1: from me, I'll let you merge this manually as discussed

ekiwi:
merged

PR #55, opened 2016-01-08
[driver] STM32: Can: use correct values to assert frequency supported

ekiwi:
Even if an unsupported frequency was used
the `prescaler > 0` assertion did not fail.

### DIFF
--- a/src/xpcc/architecture/platform/driver/can/stm32/can.hpp.in
+++ b/src/xpcc/architecture/platform/driver/can/stm32/can.hpp.in
@@ -126,7 +126,7 @@ public:
 								(SystemClock::Can{{ id }} == MHz42)?  6 : 0;
 		constexpr uint16_t prescaler = 	SystemClock::Can{{ id }} /
 								(bitrate * (1 + bs1 + bs2));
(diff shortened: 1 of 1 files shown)


ekiwi:
I will investigate, if we can make the CAN parameter selection more flexible, but right now, this fixes code that was obviously intended to function differently.

ekiwi:
I have a [python script](https://gist.github.com/ekiwi/51573729365405fef89c) that calculates the best settings for a given clock, bitrate and sampling point.
The plan is now to add a sampling point setting to the initialization (with a default value of 75%) and maybe to allow the user to specify, how well this sampling point has to be met.
As far as I have understood, the bittime needs to be perfect, assuming a perfect oscillator.
I'm just not sure how to treat the SWJ value, which will have a different effect, depending on how big BS1 and BS2 are chosen to be.

ekiwi:
Nevertheless. this commit needs to be merged first.
Any objections, @salkinium?

salkinium:
No objections, go for it.

ekiwi:
merged

PR #57, opened 2016-01-09
[driver] STM32 CAN: add CAN bit timing constants for 8MHz clock

ekiwi:
These timings have been tested in hardware with a
STM32F072 running at 8MHz and a CAN bitrate of
125kBps.


salkinium:
:+1:

PR #62, opened 2016-01-09
[driver] CAN: move bit timing calculations to separate file

ekiwi:
This is similar to how we do baudrate calculation for
UART in the stm32 driver.

As CAN seems to be very standardized, with bs1 and bs2 values
defined in the standard, the CAN calculations are placed in
generic.

This commit also adds unittests for all precalculated values,
thus we can move to calculating these values on the fly
without the risk of introducing new bugs.

### DIFF
--- a/src/xpcc/architecture/platform/driver/can/generic/can_bit_timings.hpp
+++ b/src/xpcc/architecture/platform/driver/can/generic/can_bit_timings.hpp
@@ -1,0 +1,96 @@
+// coding: utf-8
+/* Copyright (c) 2016, Roboterclub Aachen e.V.
+ * All Rights Reserved.
+ *
+ * The file is part of the xpcc library and is released under the 3-clause BSD
+ * license. See the file `LICENSE` for the full license governing this code.
+ */
+// ----------------------------------------------------------------------------
+
+#ifndef XPCC_COMMON_CAN_BIT_TIMINGS_HPP
+#define XPCC_COMMON_CAN_BIT_TIMINGS_HPP
+
+#include <xpcc/architecture/interface.hpp>
+
+#include "../../clock/generic/common_clock.hpp"
+
+namespace xpcc
+{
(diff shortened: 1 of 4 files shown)


ekiwi:
btw: the TODO will be removed, once the new algorithm is implemented.

src/xpcc/architecture/platform/driver/can/generic/can_bit_timings.hpp
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * STM32F1xx   tPCLK = 1 / 36 MHz
+ * STM32F20x   tPCLK = 1 / 30 MHz
+ * STM32F40x   tPCLK = 1 / 42 MHz
+ *
+ *
+ * @author	Kevin Laeufer
+ * @ingroup	can
+ */
+template<int32_t Clk, int32_t Bitrate, uint32_t DesiredSamplingPoint = 750000 /* 75% */>
```
salkinium: 750.000 aka. 0.750000? How accurate does this sampling point need to be?
Why not just percent aka. 75?

`DesiredSamplingPoint` is not currently used here, is that because it would be used in `calcBS{1,2}()` instead of the LUT you have now?

salkinium:
+9001 for unittests.

src/xpcc/architecture/platform/driver/can/generic/can_bit_timings.hpp
ekiwi: Yes.

I might change the DesiredSamplingPoint resolution or maybe even get rid of it and always try to archive 75%. This is still work in progress.

ekiwi:
I removed the parts that are not needed for the current implementation, such as `DesiredSamplingPoint` and rebased on top of develop.
Can I merge?

salkinium:
:+1:

PR #64, opened 2016-01-24
[driver] Timer: STM32: check for stm32f1

ekiwi:
All "modern" stm32, i.e., stm32f0, stm32f3, stm32f4
have the same behavior.
Only the stm32f1 HAL apparently declares CC1 to be
a uin16_t.


salkinium:
:+1: from me.

salkinium:
Merged manually.

ekiwi:
if you do not rebase, github will detect the merge....
So maybe we could try to merge without rebasing and if that leads to a merge commit, either go with it or start over and this time rebase.

Issue #58, opened 2016-01-09
use F_CPU to initialize xpcc::clock::fcpu_MHz and xpcc::clock::fcpu_kHz


ekiwi:
`F_CPU` is generated from the clock setting in the `project.cfg`.
Is there any reason not to use this value, to initialize `xpcc::clock::fcpu_MHz` and `xpcc::clock::fcpu_kHz`?

salkinium:
I wanted to purge `F_CPU` (and `clock` in the `project.cfg`), since it's declaration is redundant and error prone.
The `SystemClock` class knows the final CPU speed best and writes that information into the global variables.

I know that the 8MHz is not always accurate, ie. some STM32 come with a 16MHz internal clock, but `F_CPU` has the wrong semantics to use it like that.

ekiwi:
Don't we need to inject the CPU frequency for AVR microcontrollers?

salkinium:
Yes, of course only remove `clock` and `F_CPU` for Cortex-M devices (and perhaps Xmega devices).

(It might be possible to do some tricks though to get the same SystemClock interface for AVRs working, which only generates the Fuse Bit Values and places them inside a specific linker section.
This could then be interpreted by the toolchain to flash it automagically with avrdude.)

PR #66, opened 2016-01-26
Fix delay loop timing for STM32F072 / Cortex-M0

### LINKED ISSUE #65: xpcc::delayMilliseconds() inaccurate
`xpcc::delayMilliseconds(50)` on a STM32F072 (cortex-m0) using internal (HSI) or external (HSE) clock (8MHz) delays about 66ms.

salkinium:
The delay loop on the Cortex-M0 requires 4 cycles, not 3.
Fixes #65.

### DIFF
--- a/examples/stm32f072_discovery/stm32f072_discovery.hpp
+++ b/examples/stm32f072_discovery/stm32f072_discovery.hpp
@@ -87,6 +87,7 @@ switchSystemClockToHSI48()
 	if(success) {
 		xpcc::clock::fcpu_MHz = 48;
 		xpcc::clock::fcpu_kHz = 48 * 1000;
+		xpcc::clock::ns_per_loop = 4000 / 48;	// ~83ns per delay loop
 	}
 
 	return success;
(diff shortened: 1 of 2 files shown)


salkinium:
The last commit doesn't fix the [wrong default setting here though.](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/architecture/platform/driver/clock/generic/common_clock.cpp#L24) It should be 125*4 for Cortex-M0 for 8MHz.
I know we had a discussion about this before in #58.

salkinium:
Fix verified on STM32F072 and LPC11C24.

salkinium:
The reason why the loop takes 4 cycles instead of 3 is that on the Cortex-M0 the unconditional branch instruction always takes 3 cycles according to the CM0 Technical Reference Manual (TRM).

On the Cortex-M3 however: "Branches take one cycle for instruction and then pipeline reload for target instruction. Non-taken branches are 1 cycle total. Taken branches with an immediate are normally 1 cycle of pipeline reload (2 cycles total)."

Hardware is pretty cool, huh?

PR #68, opened 2016-02-02
build-system: use python2 everywhere

daniel-k:
xpcc uses a lot of custom python scripting with the assumption that `/usr/bin/env python` and the `python` executable itself point to a Python 2.x runtime. As a consequence, xpcc fails building on systems that default to a Python 3.x runtime.

This PR replaces references to `python` with `python2` everywhere in order to explicitly use a Python 2 runtime.


salkinium:
This works fine on OSX 10.9 and Ubuntu 14.04 (specifically the xpcc-vm).
Has this been tested on other platforms?

daniel-k:
For me this works on Arch Linux.

ekiwi:
in general this should work on all systems, that confirm to [PEP 0394](https://www.python.org/dev/peps/pep-0394/)

ekiwi:
tested on Fedora 23

@salkinium:  fast forward merge?

salkinium:
Yes, fast forward merge please

PR #70, opened 2016-02-04
stm32f072: really use linker script for stm32f072x8 devices

daniel-k:
The device file for stm32f072 didn't specify a linker script for ...x8 devices. I don't know if this was a mistake (or untested?) but it works now.


daniel-k:
@ekiwi 
Maybe you?

salkinium:
No, this wasn't a mistake.
The Device File Generator does not generate the `<linkerscript>` tag as I would then also have to generate 20 million tiny linkerscripts that only contain the memory sections and include one of the three main linkerscripts.

Linkerscripts should really be generated and this is a reminder that we don't currently do it.
The last time I tried to generate linkerscripts I failed at teaching SCons the dependencies on it…
Maybe @ekiwi can be our saviour?

Issue #67, opened 2016-01-27
QTCreator Support


ekiwi:
@rleh and I started playing around with [QTCreator](https://www.qt.io/ide/) yesterday.

This issue will track our progress in integrating `xpcc` with QTCreator

ekiwi:
We basically used the `Existing Project` template as documented in the [scons wiki](https://bitbucket.org/scons/scons/wiki/IDEIntegration)
We used `scons verbose=1` to see the compiler flags and then manually copied all `-D` flags to the `.config` file and  all `-I` flags to the `.includes` file. This can probably be easily automated with `scons`.
This already gives you decent auto completion and inspection (`F2`) support.

ekiwi:
QTCreator has some defines set by default, which lead the `detect.hpp` from `xpcc` to detect a hosted linux target. Thus some stuff that uses this result leads to wrong inspection results. For example the hosted implementation of `delayMilliseconds` was resolved, even though we were targeting the ARM platform.
@rleh has found out a way to work around this issue by using some additional `#define` and `#undef` in the `.config` file.

ekiwi:
Another way to resolve the above mentioned issue might be to get rid of the `detect.hpp` magic, as it is probably compiler specific and as we know the platform anyway from the device file and could propagate that information through compiler flags or through a auto generated `defines.hpp`.

rleh:
The following `#define` and `#undef` in the `.config` file lead `detect.hpp` detect an ARM platform:

```
#define XPCC__CPU_ARM
#undef XPCC__OS_UNIX
#undef __unix__
```

salkinium:
Probably not helpful for the `detect.hpp` issue, but there exists also the [QT Bare Metal plugin](http://doc.qt.io/qtcreator/creator-developing-baremetal.html). This was also presented at FOSDEM the last years.

hb128:
Thanks for your nice work :)
Could .xml files be added by default as well?

ekiwi:
> Could .xml files be added by default as well?

Unfortunately, generating the xml files is not that easy.
The `*.user` file contains user specific settings and should never be automatically overwritten which is why I do not want to generate it.
On the other hand, one should be able to creator a `.shared` file, as is briefly explained in the [QtCreator docs](http://doc.qt.io/qtcreator/creator-sharing-project-settings.html).
However, I haven't figured out how to create a `.shared` file that contains the build commands and that is accepted by QtCreator.

hb128:
[...]

PR #71, opened 2016-02-04
Basic QtCreator Project Generation Support

ekiwi:
This PR adds basic support for generating [QtCreator Generic Projects](https://github.com/qtproject/qt-creator/tree/master/src/plugins/genericprojectmanager).

I tested this quickly with our internal robot project by adding/modifying the following lines to the `SConstruct` file:

```
env.Alias('qtcreator', env.QtCreatorProject(files))
env.Alias('all', ['build', 'qtcreator', 'size'])
```

Of course, this needs some work:
- we need to add the correct work around for the `detect.hpp` issues mentioned in #67
- I want to try and generate a `.creator.shared` file that contains the build commands, which would make it possible to invoke `scons` from the QtCreator GUI
- finally GDB integration through the bare metal plugin is missing

However, I think this is already useful and therefore would like to see the PR merged, even though there are some features that should be added later.


daniel-k:
Works mostly but it doesn't inlcude `.hpp` files, so they don't appear inside QtCreator.

ekiwi:
@daniel-k @salkinium this should now include headerfiles
Can you please review my last commit?

salkinium:
This works for me, and I have to say this is pretty amazing.
Autocompletion! Even for `generated_platform`. Even for `Gpio::connect` functions.
This must be that future people have been talking about!

(The include files might want to be relative in the `.includes` file, otherwise you can't share the project files.)

ekiwi:
> The include files might want to be relative in the .includes file, otherwise you can't share the project files.

you are not supposed to share the project .... autogeneration FTW!

PR #73, opened 2016-02-05
[ui] Make line drawing symmetric

Sh4rK:
This commit makes line drawing "symmetric", which means that the pixels at both the start and end
coordinates are part of the line. Previously the "end" pixel was left out. (Which isn't always the
second coordinate, because the line drawing routine may swap them. This caused the start of the line
to shift pixels depending on where the end was.)

One consequence of this commit is that now a line is always
visible (at least 1 pixel).


salkinium:
:+1: Thanks.

salkinium:
Merged manually.

PR #78, opened 2016-02-05
Examples: Exit with EXIT_FAILURE when unrecoverable error occurred.

strongly-typed:


salkinium:
I've amended the original commit with `#include <cstdlib>` and merged manually.

PR #86, opened 2016-02-09
Use named struct instead of anonymous parameter list for drivers.

salkinium:
Via @strongly-typed:

Until now, not-so-long and long anonymous parameter lists were used to connect hardware drivers with the real hardware.

This is C-style as the order counts. If you swap Ce and Dc in the example above the code will compile and fail and it is hard to spot the error.

With a named struct as a single parameter each parameter has a distinguished name and is less error-prone. And the code is the documentation. Adding another parameter is simple, too.

The only downside is that default arguments cannot be defined.

### DIFF
--- a/examples/stm32f4_discovery/display/nokia_5110/main.cpp
+++ b/examples/stm32f4_discovery/display/nokia_5110/main.cpp
@@ -45,8 +45,15 @@ namespace lcd
 // typedef xpcc::SoftwareSpiMaster< lcd::Clk, lcd::Din> mySpiMaster;
 typedef SpiMaster2 mySpiMaster;
 
+struct Connector {
+	using Spi = mySpiMaster;
+	using Ce = lcd::Ce;
+	using Dc = lcd::Dc;
+	using Reset = lcd::Reset;
+};
+
 // create a LCD object
-xpcc::Nokia5110< mySpiMaster, lcd::Ce, lcd::Dc, lcd::Reset > display;
+xpcc::Nokia5110< Connector > display;
 
 class ThreadOne : public xpcc::pt::Protothread
 {
--- a/src/xpcc/driver/display/nokia5110.hpp
+++ b/src/xpcc/driver/display/nokia5110.hpp
@@ -25,9 +25,14 @@ namespace xpcc
  *
  * \ingroup driver_display
  */
-template < typename Spi, typename Ce, typename Dc, typename Reset >
+template < class Connector >
 class Nokia5110 : public BufferedGraphicDisplay< 84, 48 >
 {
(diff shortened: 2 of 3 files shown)


strongly-typed:
Yes, I still like my idea :-)

I would suggest to use this pattern for all new drivers and classes which use more than one parameter. It does not break compatibility.

salkinium:
The main issue I see with the approach is documentation and the naming of the types.

With the template parameters the name is contained within the driver, it essentially doesn't matter what you call it.
With this approach you have to correctly alias the types to the ones that the driver requires. If another driver calls the thing differently, you have to rename it in your code as well.
It's like having to modify your code when I rename the name of a function argument.
I find that a very awkward process.
It's also very hard to natively document with Doxygen and C++. You'd need the concept of Traits, which isn't available in C++ (yet).

However, I would like to explore this concept for abstracting the driver logic from the bus as much as possible.
We've had issues with this before, and it resulted in the classes in [`driver/bus`](https://github.com/roboterclubaachen/xpcc/tree/develop/src/xpcc/driver/bus) and the [`Lis3Transport`](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/driver/inertial/lis3_transport.hpp) classes.
In the case of the `Nokia5110` this might be a little overkill (and we already have a "bus" abstraction from `BufferedGraphicsDisplay` anyways), but perhaps there are more interesting things to discover.

salkinium:
I've been thinking more about the applications of this. I've come to really like this idea, it can be used for a lot of things.

I think the documentation and defaults problem can be solved by making the sensor provide a (documented) base type from which the user has to inherit.
This allows proper documentation and defaulting:

``` cpp
struct driver::Configuration 
{
    using SpiMaster = void; // required
    using Cs = void;    // required
    using Interrupt = xpcc::GpioUnunsed;    // optional input
    static constexpr uint8_t Revision = 2;  // driver r1 has one register less than r2
    static uint16_t crc(uint16_t input) {…}   // overwrite stateless functions like CRC
};
```

You would then inherit your custom settings and only overwrite what you care about:

``` cpp
struct DriverConfiguration : public driver::Configuration 
{
    using SpiMaster = SpiMaster3;
    using Cs = GpioOutputB4;
    // not using interrupt pin
    static constexpr uint8_t Revision = 1;  // using old revision
    // default CRC is fine
};
```

Is this a good idea? It won't be possible to inline declare the config anymore, so it can lead to more code.
But on the other hand, you won't have more than a handful of drivers in you application anyway, so I guess this is ok.

salkinium:
Partially implemented, but requires further documented policy and examples to be made best practice.

PR #89, opened 2016-02-10
[platform] Add support for STM32F103

Sh4rK:
Hi,

I've tested this in my project with manual clock configuration, and it works as expected. I took the device description files from the experimental branch, but removed the clocktree as that's not used in develop, and missing from the rest of the devices too. I haven't yet added an example for this, but if needed, I can probably write a blinky.

Just for reference, in my project I configured the clock like this for 32MHz:

``` c++
ClockControl::enableInternalClock();
ClockControl::enablePll(ClockControl::PllSource::InternalClock, ClockControl::UsbPrescaler::Div1, 8);
ClockControl::enableSystemClock(ClockControl::SystemClockSource::Pll);
ClockControl::setAhbPrescaler(ClockControl::AhbPrescaler::Div1);
ClockControl::setApb1Prescaler(ClockControl::Apb1Prescaler::Div1);
ClockControl::setApb2Prescaler(ClockControl::Apb2Prescaler::Div1);
xpcc::clock::fcpu        = systemClock::Frequency;
xpcc::clock::fcpu_kHz    = systemClock::Frequency / 1000;
xpcc::clock::fcpu_MHz    = systemClock::Frequency / 1000000;
xpcc::clock::ns_per_loop = std::round(1000000000.f / systemClock::Frequency * 3.f);
```

And here's the dummy clock:

``` c++
// Dummy clock for devices
struct systemClock {
    static constexpr uint32_t Frequency = 32000000;
    static constexpr uint32_t Usart1 = Frequency;
    static constexpr uint32_t Can1 = Frequency;
    static constexpr uint32_t I2c2 = Frequency;
};
```

Also a note: `ClockControl::PllSource::InternalClock` should probably be renamed to something like `ClockControl::PllSource::InternalClockDiv2`. It confused me when I set the multiplier to 4, and I didn't get 32MHz.

### DIFF
--- a/src/xpcc/architecture/platform/devices/stm32/stm32f103-c_r_t_v-8_b.xml
+++ b/src/xpcc/architecture/platform/devices/stm32/stm32f103-c_r_t_v-8_b.xml
@@ -1,0 +1,388 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!DOCTYPE rca SYSTEM "../devicefile.dtd">
+<rca version="1.0">
+  <!-- WARNING: This file is generated automatically, do not edit!
+ 		Please modify the xpcc/tools/device_file_generator code instead and rebuild this file.
+ 		Be aware, that regenerated files might have a different composition due to technical reasons. -->
+  <device platform="stm32" family="f1" name="103" pin_id="c|r|t|v" size_id="8|b">
+    <flash>65536</flash>
+    <ram>20480</ram>
+    <core>cortex-m3</core>
+    <linkerscript device-size-id="8">stm32f103_8.ld</linkerscript>
+    <linkerscript device-size-id="b">stm32f103_b.ld</linkerscript>
+    <pin-count device-pin-id="v">100</pin-count>
+    <pin-count device-pin-id="t">36</pin-count>
+    <pin-count device-pin-id="c">48</pin-count>
+    <pin-count device-pin-id="r">64</pin-count>
+    <header>stm32f1xx.h</header>
+    <define>STM32F103xB</define>
+    <driver type="core" name="cortex">
+      <memory access="rx" start="0x8000000" name="flash" size="64"/>
+      <memory access="rwx" start="0x20000000" name="sram1" size="20"/>
+      <vector position="0" name="WWDG"/>
+      <vector position="1" name="PVD"/>
+      <vector position="2" name="TAMPER"/>
+      <vector position="3" name="RTC"/>
+      <vector position="4" name="FLASH"/>
+      <vector position="5" name="RCC"/>
+      <vector position="6" name="EXTI0"/>
+      <vector position="7" name="EXTI1"/>
+      <vector position="8" name="EXTI2"/>
+      <vector position="9" name="EXTI3"/>
+      <vector position="10" name="EXTI4"/>
+      <vector position="11" name="DMA1_Channel1"/>
+      <vector position="12" name="DMA1_Channel2"/>
+      <vector position="13" name="DMA1_Channel3"/>
+      <vector position="14" name="DMA1_Channel4"/>
+      <vector position="15" name="DMA1_Channel5"/>
+      <vector position="16" name="DMA1_Channel6"/>
+      <vector position="17" name="DMA1_Channel7"/>
+      <vector position="18" name="ADC1_2"/>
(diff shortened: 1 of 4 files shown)


ekiwi:
@salkinium can you check, if the two device files could be merged?

salkinium:
@ekiwi I've tried that before, didn't go so well.

The devices are merged manually, [first by size, then by name](https://github.com/roboterclubaachen/xpcc/blob/develop/tools/device_file_generator/merger.py?ts=4#L39-L40).
The heuristics for the STM32F1 series are as follows:
1. [merging by size](https://github.com/roboterclubaachen/xpcc/blob/develop/tools/device_file_generator/merger.py?ts=4#L393-L399), then
2. [merging by name](https://github.com/roboterclubaachen/xpcc/blob/develop/tools/device_file_generator/merger.py?ts=4#L128-L132).

As you can see, I chose to follow ST's nomenclature with the sizes correlating to F1 "density".
This does make some sense, since the F1 has grown organically (it was ST's first Cortex-M series) and therefore has a lot of quirks like having their features dependent on the "size" id.

Merging all F103 devices together [results in this, which is actually quite horrible](https://gist.github.com/salkinium/818fdedc3f416a02af6b) (look at the vector table, then very cautiously  work yourself downwards).

I mean, ST made this pretty clear with their defines:

``` xml
    <define device-size-id="4|6" device-pin-id="c|r|t">STM32F103x6</define>
    <define device-size-id="8|b" device-pin-id="c|r|t|v">STM32F103xB</define>
    <define device-size-id="c|d|e" device-pin-id="r|v|z">STM32F103xE</define>
    <define device-size-id="f|g" device-pin-id="r|v|z">STM32F103xG</define>
```

I would be sort of ok with [merging the medium and high density devices](https://gist.github.com/salkinium/57c26196d1698366600c) (ie. the two device files in this PR).
The results are not exactly great though, and you would still have the low and super high densities separately ('cos **NO**).

salkinium:
> I haven't yet added an example for this, but if needed, I can probably write a blinky.

Only for the Nucleo-F103, because there is little point in having examples for hardware that we don't have ;-)
Adding an example is also the only way to add the STM32F103 to our CI setup, which will then make it a little easier to not break support for the F103 in the future.
It's optional though, I can add it if you don't want to (I already have another Nucleo-32 example locally anyways).

> Also a note: `ClockControl::PllSource::InternalClock` should probably be renamed to something like `ClockControl::PllSource::InternalClockDiv2`. It confused me when I set the multiplier to 4, and I didn't get 32MHz.

Interesting usability feedback. The same then applies to [`ClockControl::ClockOutputSource::Pll`](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/architecture/platform/driver/clock/stm32/clock.hpp.in?ts=4#L147) which should be `PllDiv2` instead, since it's also divided by 2.
I'll have a look at how common these dividers are across all devices and then change it accordingly. 

Thanks!

src/xpcc/architecture/platform/devices/stm32/stm32f103-c_r_t_v-8_b.xml
```diff
@@ -0,0 +1,388 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!DOCTYPE rca SYSTEM "../devicefile.dtd">
+<rca version="1.0">
+  <!-- WARNING: This file is generated automatically, do not edit!
+ 		Please modify the xpcc/tools/device_file_generator code instead and rebuild this file.
+ 		Be aware, that regenerated files might have a different composition due to technical reasons. -->
+  <device platform="stm32" family="f1" name="103" pin_id="c|r|t|v" size_id="8|b">
+    <flash>65536</flash>
```
salkinium, line 8: Hm, there is a bug here, the `B` size has 128kB of flash.
I was starting to wonder why the two sizes are identical in this file.
I'm looking in the device file generator for a parsing error.

salkinium, line 8: Yeah, found a bug in the stm32 device file reader. It will be fixed in a separate commit though.

Sh4rK:
> It's optional though, I can add it if you don't want to

I'd thank you for it if you do :)

PR #90, opened 2016-02-11
[build] Fix building xpcc as a dependency

Sh4rK:
This fixes building xpcc as a dependency (e.g. xpcc is a subdirectory of the project). Previously the library didn't even build, and the path was also wrong.

### DIFF
--- a/scons/site_tools/xpcc.py
+++ b/scons/site_tools/xpcc.py
@@ -150,8 +150,8 @@ def xpcc_library(env, buildpath=None):
 								  'templates/xpcc_config.hpp.in'),
 			substitutions = substitutions)
 
-	env.PrependUnique(LIBS = ['xpcc'])
-	env.PrependUnique(LIBPATH = [os.path.join(env['XPCC_BUILDPATH'], 'src')])
+	env.PrependUnique(LIBS = [library])
+	env.PrependUnique(LIBPATH = [os.path.join(env['XPCC_BUILDPATH'], 'src', 'xpcc', 'src')])
 
 	# restore original environment
 	env['XPCC_BUILDPATH'], env['XPCC_BASEPATH'] = backup


ekiwi:
@Sh4rK Could you please elaborate on why this is needed?
I don't really understand.

Sh4rK:
> Could you please elaborate on why this is needed?

Because the build didn't work before at all, and this fixes it.

Try building something in this directory configuration:

```
root/
  src/
  xpcc/
  SConstruct
  project.cfg
```

where SConstruct looks something like this:

``` python
rootpath = 'xpcc'

env = Environment(tools = ['xpcc'], toolpath = [rootpath + '/scons/site_tools'])

# find all source files
files = env.FindFiles('src')

# build the xpcc library
env.XpccLibrary()

# build the program
program = env.Program(target = env['XPCC_CONFIG']['general']['name'], source = files.sources)

# create a file called 'defines.hpp' with all preprocessor defines if necessary
env.Defines()

env.Alias('size', env.Size(program))
env.Alias('symbols', env.Symbols(program))
env.Alias('defines', env.ShowDefines())

hexfile = env.Hex(program)

env.Alias('build', [hexfile, env.Listing(program)])
env.Alias('all', ['build', 'size'])

env.Default('all')
```

ekiwi:
Thank you for the explanation.

> Because the build didn't work before at all, and this fixes it.

Could you show us the error that you get?
If you run `scons verbose=1` this would be helpful, as it shows all flags that are handed to gcc.

Sh4rK:
The xpcc library doesn't build at all, and then when linking the program it complains that it can't find `-lxpcc`.

Here's the output: https://gist.github.com/Sh4rK/442bda6663aeed890b6e

ekiwi:
Thank you for the output you provided. I will be busy until Tuesday, then I'll investigate the issue, as I want to understand, why this is happening, before coming up with a solution.

In the mean time: could you also provide me with the output of running `scons --tree=all`?
It seems that there is some issue with the dependencies.

Sh4rK:
Here's the output with `--tree-all`:
https://gist.github.com/Sh4rK/99b7a63dcaa30f983145

salkinium:
@ekiwi Any update on this?

PR #91, opened 2016-02-12
Add Nucleo-F103RB blinky example

salkinium:
Adds a blinky example for the NUCLEO-F103RB which blinks the LED and prints a counter in a loop.

@Sh4rK: I ran into the same problem you had with higher clock frequencies. It turns out that the `ClockControl` code does not set the Flash Latency correctly, this has to be done manually. [For >48MHz  you need 2 wait states.](https://github.com/roboterclubaachen/xpcc/blob/develop/examples/nucleo_f103rb/nucleo_f103rb.hpp?ts=4#L104)
This is a stupid design flaw of `ClockControl`, I'll have to change that.

### DIFF
--- a/examples/nucleo_f103rb/nucleo_f103rb.hpp
+++ b/examples/nucleo_f103rb/nucleo_f103rb.hpp
@@ -1,0 +1,133 @@
+/* Copyright (c) 2016, Roboterclub Aachen e.V.
+ * All Rights Reserved.
+ *
+ * The file is part of the xpcc library and is released under the 3-clause BSD
+ * license. See the file `LICENSE` for the full license governing this code.
+ * ------------------------------------------------------------------------ */
+
+//
+// NUCLEO-F103RB
+// Nucleo kit for STM32F103RB
+// http://www.st.com/web/catalog/tools/FM116/SC959/SS1532/LN1847/PF259875
+//
+
+#ifndef XPCC_STM32_NUCLEO_F103RB_HPP
+#define XPCC_STM32_NUCLEO_F103RB_HPP
+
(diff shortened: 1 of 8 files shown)


examples/nucleo_f103rb/nucleo_f103rb.hpp
```diff
@@ -0,0 +1,133 @@
...
+	ClockControl::setApb2Prescaler(ClockControl::Apb2Prescaler::Div1);
+	// update frequencies for busy-wait delay functions
+	xpcc::clock::fcpu        = systemClock::Frequency;
+	xpcc::clock::fcpu_kHz    = systemClock::Frequency / 1000;
+	xpcc::clock::fcpu_MHz    = systemClock::Frequency / 1000000;
+	xpcc::clock::ns_per_loop = std::round(1000000000.f / systemClock::Frequency * 3.f);
+
+	xpcc::cortex::SysTickTimer::initialize<systemClock>();
+
+	stlink::Rx::connect(stlink::Uart::Tx);
```
salkinium: This doesn't look right.

ekiwi:
> It turns out that the ClockControl code does not set the Flash Latency correctly

I am sorry, I forgot: [code from 2013](https://github.com/salkinium/xpcc/blob/7303e5a3ddb3b47b5b457de89f1ba1fd94d1b9ee/src/xpcc/architecture/platform/driver/core/cortex/stm32/stm32.macros#L63)

salkinium:
@ekiwi This is also the reason why the `HSI48` on the STM32F072 discovery doesn't work correctly. It requires 1 wait stage preferably with prefetch buffer enabled.

PR #96, opened 2016-02-18
Add Nucleo-F429ZI blinky example

salkinium:
Adds a board config with example for the Nucleo-144 F429ZI target.

cc @strongly-typed: I couldn't get the board programmed, due to what I think is a bug in the STLink firmware. Can you try this?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,6 +62,7 @@ env:
  - TEST_SUITE="check=examples examples=stm32f469_discovery"
  - TEST_SUITE="check=examples examples=nucleo_f103rb"
  - TEST_SUITE="check=examples examples=nucleo_f411re"
+ - TEST_SUITE="check=examples examples=nucleo_f429zi"
  - TEST_SUITE="check=examples examples=unittest"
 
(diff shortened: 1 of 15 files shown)


salkinium:
This is on hold, since I cannot program the NUCLEO-F429ZI on OS X v10.9.5. Something to do with OSX USB implementation.

strongly-typed:
Same problem here with OS X 10.11.3.

There has always been a workaround in place

```
 https://github.com/ntfreak/openocd/blob/master/src/jtag/drivers/stlink_usb.c#L1622
```

There are rumors that OS X 10.11 changed something so that `jtag_libusb_reset_device` does not work anymore.

```
http://libusb.6.n5.nabble.com/libusb-libusb-reset-device-not-working-on-Mac-OS-X-10-11-El-Capitan-td5715251.html
```

I can program the board on the first launch of `openocd`. So now I never close `openocd` and manually flash the board.

@salkinium But there are other reasons why you should not work so much on that issue now ;-)

salkinium:
Updating the ST-Link firmware to the latest version allows for normal programming.

PR #97, opened 2016-02-21
Add heap sections to linkerscript and better memory usage printout

salkinium:
Since the stack memory usage is now bound we can:
1. clarify the `.stack` section memory usage to make it clearer how much static memory is used.
2. accurately calculate the available heap size, since the stack now has a fixed size.
3. remove  `.vectors` from the flash sections, since it is now completely overlaid on the `.reset` section.

This PR adds heap section information and improves the memory usage printout.
It also corrects a mistake in the Device Files, where the CCM size was not part of the top-level `ram` tag and therefore the usage of RAM in percent was higher than actual.
### Memory usage printout

```
Device: stm32f407vg

Program:    2400 bytes (0.2% Full)
(.data + .fastdata + .reset + .rodata + .text + .vectors)

Data:       3740 bytes (2.9% Full)
(.bss + .data + .fastdata + .noinit + .stack + .vectors)
```

The new printout is much more verbose and accurate:

```
Device: stm32f407vg

Program:    2012 bytes (0.2% used)
(.data + .fastdata + .reset + .rodata + .text)

Data:       3740 bytes (1.9% used) = 412 bytes static (0.2%) + 3328 bytes stack (1.7%)
(.bss + .data + .fastdata + .noinit + .stack + .vectors)

Heap:     130668 bytes (66.5% available)
(.heap1 + .heap2 + .heap3)
```

(Note: the 412 bytes static usage are actually only 8 bytes + vector table in RAM).
### Heap sections

Five new memory sections are added to the linker scripts:
- `.heap0` is free memory in non-executable CCM (F4) or DTCM (F7)
- `.heap1` is free memory in SRAM1 (all STM32)
- `.heap2` is free memory in SRAM2 (F3, F4, F7)
- `.heap3` is free memory in SRAM3 (F4)
- `.heap4` is free memory in executable CCM (F3) or ITCM (F7)

Note that Heap and Data usage do not always add up to 100% RAM, because heap allocation currently can only allocate within one continous memory section (`.heap1` to `.heap2` or `.heap3`), and therefore cannot make use of the free memory in .heap0 (CCM) and .heap4 (ITCM).
This would require an allocator with support for multiple traits (DMA-able, fast access, etc.) and multiple non-continous sections.

### DIFF
--- a/scons/site_tools/helper.py
+++ b/scons/site_tools/helper.py
@@ -52,11 +52,11 @@ import subprocess
 filter = re.compile('^(?P<section>[.]\w+)\s*(?P<size>\d+)\s*(?P<addr>\d+)$')
 
 # Sections which will remain in the Flash
-flashSectionNames = ['.reset', '.vectors', '.fastcode', '.fastdata', '.text', '.rodata', '.data']
+flashSectionNames = ['.reset', '.fastcode', '.fastdata', '.text', '.rodata', '.data']
 
 # Sections which will be created in RAM or are copied from the Flash. In that
 # case the section will appear also in `flashSectionNames`.
-ramSectionNames = ['.vectors', '.fastcode', '.fastdata', '.data', '.bss', '.noinit', '.stack']
+ramSectionNames = ['.vectors', '.fastcode', '.fastdata', '.data', '.bss', '.noinit']
 
 def size_action(target, source, env):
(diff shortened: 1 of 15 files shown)


salkinium:
Note to self @salkinium: https://github.com/cbiffle/etl/blob/master/mem/arena.h for non-continous memory sections allocation.

PR #98, opened 2016-02-24
[io] IOStream: printf: Fix the position of the minus sign.

strongly-typed:
In xpcc printf the minus does not touch the number.
glibc's printf does and is (to my experience) the most common behaviour.

Info:    xpcc  >>-  42<<
Info:    glibc >>  -42<<

With this PR xpcc's printf matches the output of glibc's printf:

Info:    xpcc  >>  -42<<
Info:    glibc >>  -42<<

I know that the printf is not the best available printf for STM32 but no-one else implemented a better printf for xpcc yet.

### DIFF
--- a/examples/linux/printf/SConstruct
+++ b/examples/linux/printf/SConstruct
@@ -1,0 +1,33 @@
+# path to the xpcc root directory
+rootpath = '../../..'
+
+env = Environment(tools = ['xpcc'], toolpath = [rootpath + '/scons/site_tools'])
+
+# find all source files
+files = env.FindFiles('.')
+
+# build the program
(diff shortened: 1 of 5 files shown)


salkinium:
:+1:

> I know that the printf is not the best available printf for STM32 but no-one else implemented a better printf for xpcc yet.

Well, the entire IOStream isn't the most efficient design possible, but it works and it is resonably usable.

I'm a bit bummed though, that we don't have unit tests for our `printf` implementation.
For the IOStream [the unit tests covers negative number printing](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/io/test/io_stream_test.cpp#L110).
Hmpf.

Issue #26, opened 2015-07-18
WriteFloat implementation for type double on ARM Targets


TheTh0r:
Double currently is casted to float and the writeFloat function is used in the IOStream.

(See TODO in /src/xpcc/io/iostream_float.cpp line 93)

salkinium:
And floats cannot be formatted by number of decimal points or similar, because it's all hard-coded.

Also interesting and mildly related to the topic is an alternative implementation of the OStream:
http://probablydance.com/2014/08/30/format_it-iterator-based-string-formatting/

PR #99, opened 2016-02-24
[io] IOStream: printf now supports basic %m.nf float formatting.

### LINKED ISSUE #18: IOStream cannot format floats
This is actually a special case of: We need a Number Formatter.

More specifically, we need a Number Formatter that implements the format patterns from the [Unicode Technical Standard #35](http://www.unicode.org/reports/tr35/tr35-numbers.html#Number_Format_Patterns).

strongly-typed:
Simple %m.nf now gives a result quite close to what glibc does.

Fixes #18.

### DIFF
--- a/examples/linux/printf/SConstruct
+++ b/examples/linux/printf/SConstruct
@@ -1,0 +1,33 @@
+# path to the xpcc root directory
+rootpath = '../../..'
+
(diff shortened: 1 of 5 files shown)


salkinium:
If the added program size for `printf` with float formatting proves too much for some AVRs, we'll make it configurable then.
If you use printf in AVRs there is a large overhead anyways, so this isn't news.
:+1: from me.

Thanks. Also related to #26.

PR #102, opened 2016-02-28
Debugger-friendly hard fault handler

salkinium:
The hard fault handler moves the stack pointer which disallows a connected debugger to do a backtrace and gather valid stack frames.

Therefore, if a debugger is connected (as signalled by the `C_DEBUGEN` bit in the `CoreDebug->DHCSR` register) a breakpoint it triggered _before_ continuing executing the hard fault handler.
This allows debugging of the controller without recompiling or reflashing which potentially preserves the hard fault reason (recompiling can move things around). A reset is however still required!

OpenOCD does not automatically reset the `C_DEBUGEN` bit on the `shutdown` command, so all `project.cfg` using OpenOCD had to be adapted to force this.
I'm not sure if this is a bug, or if this is expected behaviour. It could also be that no-one has ever used it this way before.

This PR adds the following:
- breakpoint triggering in hard fault handler if `C_DEBUGEN` is set.
- traditional busy waiting in hard fault handler if neither LED nor Log is enabled.
- forcing the `C_DEBUGEN` bit low after flashing using openocd.
- Example for running this on the STM32F4.

cc @ekiwi @daniel-k:
Please test the hard fault handler on STM32F0, since for the Cortex-M0, software access to `CoreDebug->DHCSR` is implementation defined, and for LPC11 it is not implemented (but I believe for F0 it is).

### DIFF
--- a/examples/stm32f4_discovery/hard_fault/SConstruct
+++ b/examples/stm32f4_discovery/hard_fault/SConstruct
@@ -1,0 +1,4 @@
+# path to the xpcc root directory
+xpccpath = '../../..'
+# execute the common SConstruct file
+execfile(xpccpath + '/scons/SConstruct')
--- a/examples/stm32f4_discovery/hard_fault/main.cpp
+++ b/examples/stm32f4_discovery/hard_fault/main.cpp
@@ -1,0 +1,72 @@
+#include "../stm32f4_discovery.hpp"
+#include <xpcc/debug/logger.hpp>
+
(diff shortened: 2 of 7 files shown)


salkinium:
> Please test the hard fault handler on STM32F0, since for the Cortex-M0, software access to CoreDebug->DHCSR is implementation defined, and for LPC11 it is not implemented (but I believe for F0 it is).

Software access to `CoreDebug->DHCSR` is not implemented in STM32F0 either. Doh.

PR #103, opened 2016-03-06
Add universal SConstruct file and deduplicate project.cfg data

salkinium:
I wanted to improve our "Getting Started" guide and I realized how terrible everything is.
We've been copying the same `Sconstruct` file over and over around and I'm sick of it. Also the `project.cfg` are way to redundant.
- I've added a universal `SConstruct` for `avr`, `cortex-m` and `hosted`. It works with extra sources and XPCC communication and images.
- Adapted all `SConstruct` files to delegate to this file. Geez.
- GDB tool is now available to all Cortex-M programs. Usage: `scons openocd-debug`, then `scons debug`, which opens `arm-none-eabi-gdb -tui` with the hex file.  Would be nice to only have to do `scons debug` and the openocd server starts automatically and stops after GDB has quit.
- Purge the last of `CORTEX_VECTORS_RAM`.

Additionally I removed some nonsense from the `project.cfg` syntax:
- Remove leading newline, because WTF?
- Remove `[program]` section, which contained `tool = openocd`, only to be followed by the `[openocd]` sections. Btw, for AVRs there is no `program.tool = avrdude`, because the whole thing is redundant.
- Remove all redundant `openocd.cfg` files. They were also just copied around, and most of them were overly verbose. They can been inlined, since OpenOCD looks in all its search paths for the file. So `openocd.configfile = board/stm32f4discovery.cfg` is the way to go.
- Default and remove redundant `openocd.commands`. This is literally always the same.
- Adapt build paths so that they don't overwrite each other.
- What does `scons.regenerate = false` do differently than the default setting?!?

Please note that all of these changes are backwards compatible.
You can still you your own `SConstruct` file, you can still use your own `openocd.cfg` file.
The reduction of verbosity in `project.cfg` does not break existing code.

Example `SConstruct` file:

``` python
xpccpath = '../../..'
execfile(xpccpath + '/scons/SConstruct')
```

Areas of interest for review:
- is `xpcc/scons/SConstruct` a good location to put this common code?
- is the use of `execfile` ok?
- naming things in `project.cfg`:
  - `build.extrasources`?
  - `communication.source`, `communication.generate`?
  - `images.source`?
- building `GitInfoHeader` and `BuildInfoHeader` always.
- using `arm-none-eabi-gdb` always with `-tui` option.
- Debugging example updated.

@ekiwi @dergraaf @strongly-typed please review.


ekiwi:
Thank you Niklas, great work!

While we are at it, I would like to  make some suggestions.

Should we change the default build path in the `project.cfg`  to `build/`?
This makes it much more transparent to the user how the build process works and makes it easier to hand the `.elf` to `gdb`, in my oppinion.

ekiwi:
> Would that make it possible to set the dependencies in the right order for linkerscript generation?

If I remember correctly, yes.

salkinium:
> > Would that make it possible to set the dependencies in the right order for linkerscript generation?
> 
> If I remember correctly, yes.

:heart: :tada:

scons/SConstruct
```diff
@@ -0,0 +1,85 @@
...
+        env.Alias('program', env.LpcLinkFlash(program))
+        env.Alias('lpclink-debug', env.LpcLinkDebug(program))
+    # ST Link only reasonable on Windows, please use OpenOCD instead
+    if env.CheckArchitecture('avr'):
+        env.Alias('eeprom', env.Eeprom(program))
+    else:
+        env.Alias('debug', env.GdbDebug(program))
+        env.Alias('bin', env.Bin(program))
+
+    env.Alias('listing', env.Listing(program))
```
salkinium, line 82: @ekiwi I made the `.lss` and `.bin` file creation optional, so that when you run scons check, it consumes a little less space.

ekiwi, line 82: Great! This probably saves about 500MB!

salkinium:
> Should we change the default build path in the project.cfg to build/?

Hm, I see the issue from a user perspective, because the build folder is not where you'd expect it.
I like that I can do `rm -r xpcc/build` to clean up _all_ of the build files after a `scons check`, so I would be inclined to leave this as it is and make this point clear in the "Getting Started" guide.
(`./build` is [the default path](https://github.com/salkinium/xpcc/blob/feature/fix_sconstruct_project_files/scons/site_tools/xpcc.py#L337) already)

PR #104, opened 2016-03-07
Replace MAIN_FUNCTION with xpcc_main()

salkinium:
This deprecates `MAIN_FUNCTION` in favour of a much saner `xpcc_main()` macro, which looks like a lot less black magic:

``` cpp
int xpcc_main() {...}
// is much better than
MAIN_FUNCTION {...}
```

Additionally, for hosted, the main function arguments can now transparently be added:

``` cpp
int xpcc_main(int argc, int** argv) {...}
// arguments can also be renamed, like a normal function
int xpcc_main(int foo, int** bar) {...}
// or to avoid unused parameter warnings
int xpcc_main(int, int**) {...}
```

If this signature is compiled for embedded targets, the function arguments are dismissed entirely.

This change is backwards compatible, however, I couldn't find a way to add a deprecation warning whenever `MAIN_FUNCTION` is still used.

cc @ekiwi @dergraaf objections?


salkinium:
Another important point here is that this macro is namespaced with `xpcc_`, since `MAIN_FUNCTION` is way too generic.

I'd also like to namespace the following:

| Current | Proposal |
| :-- | :-- |
| `ALWAYS_INLINE` | `xpcc__always_inline` |
| `ATTRIBUTE_UNUSED` | `xpcc__unused` |
| `ATTRIBUTE_WEAK` | `xpcc__weak` |
| `ATTRIBUTE_ALIGNED(n)` | `xpcc__aligned(n)` |
| `ATTRIBUTE_PACKED` | `xpcc__packed` |
| `ATTRIBUTE_FASTCODE` | `xpcc__fastcode` |
| `ATTRIBUTE_FASTDATA` | `xpcc__fastdata` |
| `ATTRIBUTE_MAY_ALIAS` | `xpcc__may_alias` |
| `CONCAT(a,b)` | `XPCC__CONCAT(a,b)` |
| `STRINGIFY(a)` | `XPCC__STRINGIFY(a)` |
| `ENUM_CLASS_FLAG` | :fire: :fire: :fire: |

Alternative prefix for compiler attributes could be `__xpcc__always_inline` or `__xpcc_always_inline`?

dergraaf:
It might even be worth considering removing the special handling of main altogether. It save two bytes of RAM for AVRs and adds nothing for the other targets. If the two bytes are really important for a project, the attribute can be added by hand.

Concerning the second proposal: Names containing a double underscore are reserved for the standard library and compiler implementation. The same applies for names beginning with an underscore.

A good overview can be found here: http://stackoverflow.com/a/228797

salkinium:
> It might even be worth considering removing the special handling of main altogether.

The whole "hosted/embedded same code compilation" is broken anyways, unless you abstract the hardware explicitly like we do in the robots.
And then you still have to check `#ifdef XPCC__OS_HOSTED` and do special things then.
So I agree, this should just be removed.

Regarding underscores: Is that really such a big deal? I doubt `__xpcc_` will be used anywhere.
Btw, [here is "that other" OS's attributes abstraction](https://github.com/armmbed/compiler-polyfill).

dergraaf:
> Regarding underscores: Is that really such a big deal? I doubt __xpcc_ will be used anywhere.

That is not the point. Why doing it wrong when it is just as easy to do it right. What do we gain from the underscores?
I have to admit that I'm responsible for introducing the double underscore in xpcc. I don't think it is necessary to remove all of them at once, but for new stuff we should stick to the standard.

salkinium:
> I don't think it is necessary to remove all of them at once, but for new stuff we should stick to the standard.

Ok, would the `xpcc_always_inline` scheme be ok then?

dergraaf:
> Ok, would the xpcc_always_inline scheme be ok then?

No objections from my side :-)

salkinium:
Closing this to archive this discussion and alternate implementation.
Finalized in #105.

PR #109, opened 2016-03-11
Add board targets

salkinium:
This allows a radical reduction in `project.cfg` lorem ipsum, where you can now declare a board as a target:

``` ini
[build]
board = stm32f4_discovery
```

`xpcc.py` will then search in `xpcc/architecture/platform/board` for the board and augment the `project.cfg` with the `board.cfg`, which contains the rest of the information.
Some hacking had to be done in the Scanner in `configfile.py`, which has to "find" source files inside the board folders, even though they are ignored.

This backwards compatible change moves all boards to the new location and adapts all examples.

cc @ekiwi @dergraaf: Have a look at the changes in https://github.com/salkinium/xpcc/commit/d5f608f29149ab5c35c7a6452d23b8846db337bf please.


ekiwi:
If you move the import and Travis CI passes, I will merge this commit. OK?

salkinium:
> If you move the import and Travis CI passes, I will merge this commit. OK?

Hang on, I have some more commits to add.

salkinium:
I added some minor changes, most notably a `typedef -> using` refactoring and adding the `serialStream`  and the `xpcc::Clock` code to Arduino Uno board target.

I also grouped the board LEDs together as SoftwarePort, which is used by a generic blinky example that I will open a PR for later. It cycles through the LEDs forwards or backwards depending on Button press:

``` cpp
#include <xpcc/architecture/platform.hpp>
int main()
{
    Board::initialize();
    Board::Leds::setOutput();
    const uint16_t max = (1 << (xpcc::max(Board::Leds::width, uint8_t(2)) - 1));
    uint16_t value(1);

    while (1) {
        if (Button::read()) {
            value >>= 1;
            if (value == 0) value = max;
        }
        else {
            value <<= 1;
            if (value > max) value = 1;
        }
        Board::Leds::write(value);
        xpcc::delayMilliseconds(250);
    }
    return 0;
}
```

That example is actually the reason for this PR, I needed a way for the Getting Started guide to be generic and simple enough for it to be comprehensible.

> > If you move the import and Travis CI passes, I will merge this commit. OK?
> 
> Hang on, I have some more commits to add.

Should we use this semantic? The "in progress" label cannot be attached by contributors after all.
https://github.com/sanemat/do-not-merge-wip-for-github

salkinium:
I will rebase and squash the commits after your review,  please then merge `--no-ff` because major changes.

ekiwi:
> Should we use this semantic? The "in progress" label cannot be attached by contributors after all.
> https://github.com/sanemat/do-not-merge-wip-for-github

I like this idea, but let's use `[rfc]` instead of `[wip]`, as this specifies the reason why you are posting your patch set even though you do not want it to be merged at the moment.

salkinium:
> I like this idea, but let's use [rfc] instead of [wip], as this specifies the reason why you are posting your patch set even though you do not want it to be merged at the moment.

But isn't the `[rfc]` part obvious from posting it as a PR?
The `[wip]` allows the contributor to get some feedback before completing the entire feature and having the reviewer prematurely merge the PR.

PR #111, opened 2016-03-13
Generate linkerscripts using Jinja2 template

salkinium:
This change adds linkerscript generation from the information inside the device files.
This is merely a relatvely simple change, but it removes the last hurdle to support all STM32 devices! Huray! \o/

This also adds the ability to modify the allocated stack size (default is 2kB - 192B):

``` ini
[parameters]
core.cortex.0.main_stack_size = 3000   # mooaaar rraaaaaaammmmm
```

cc @dergraaf  @ekiwi: This is a breaking change for the robot code, since the `env.XpccLibrary()` call needs to be moved!


src/xpcc/architecture/platform/driver/core/cortex/driver.xml
```diff
@@ -16,11 +17,23 @@
...
+		<!-- lpc -->
+		<template device-platform="lpc" out="linkerscript.ld">stm32/stm32_ram.ld.in</template>
+		<!-- stm32f0, stm32f1, stm32f2 -->
+		<template device-family="f0|f1|f2" out="linkerscript.ld">stm32/stm32_ram.ld.in</template>
+		<!-- stm32f3 -->
+		<template device-family="f3" device-name="303|328|334|358|398" out="linkerscript.ld">stm32/stm32_iccm.ld.in</template>
+		<template device-family="f3" device-name="301|302|318|378|373" out="linkerscript.ld">stm32/stm32_ram.ld.in</template>
+		<!-- stm32f4 -->
+		<template device-family="f4" device-name="405|407|415|417|427|429|437|439|469|479" out="linkerscript.ld">stm32/stm32_dccm.ld.in</template>
+		<template device-family="f4" device-name="401|410|411|446" out="linkerscript.ld">stm32/stm32_ram.ld.in</template>
```
salkinium, line 33: @dergraaf At this stage I really wanted to be able to look at the content of the device file to check if it contains the `ccm` memory section. I came to the conclusion that this file should be replaced by a python class which builds this together dynamically. Then I realized we really need a module system :-(

salkinium:
> This is a breaking change for the robot code, since the `env.XpccLibrary()` call needs to be moved!

There are 39 calls to `env.XpccLibrary()` in the wrong order in the robot code. Argh, whyyyyy?

ekiwi:
> This is a breaking change for the robot code, since the env.XpccLibrary() call needs to be moved!

This change actually most likely breaks **every** existing xpcc project as they all copied the default `SConstruct`.
I don't want to say that this is a reason to not merge, it is just something that we need to be aware of.

salkinium:
> This change actually most likely breaks every existing xpcc project as they all copied the default SConstruct.

Good point.
Maybe I can detect this somehow and create a warning.

salkinium:
Hm, interestingly the linkerscript generation works fine if I put the `env.XpccLibrary()` after `env.Program()`.
I think the "manual" `Depends` on the `.elf` file works regardless.

ekiwi:
> Maybe I can detect this somehow and create a warning.

Yes, this should be possible. When I have found my code .....
My solution hooks into the env.Program emitter and declares the linkerscript dependency there. We can check if the LINKERSCRIPT environment variable has been set. If it is missing we could emit a custom error message.

salkinium:
I've tested this in hardware on all STM32 boards we support as well as the LPCxpresso.

salkinium:
> When I have found my code .....

@ekiwi: Found your code yet?

Issue #112, opened 2016-03-14
Hard fault handler does not compile without optimization


rleh:
Disabling compiler optimization (`"-Os"` -> `"-O0"` in `scons/site_tools/arm.py:117`) produces the following error:

``` bash
$ cd examples/stm32f4_discovery/hard_fault
$ scons
/* ... */
Compiling C++: /home/user/xpcc/build/stm32f4_discovery/hard_fault/libxpcc/build/stm32f4_discovery/hard_fault/libxpcc/generated_platform/driver/core/cortex/hard_fault_handler.o
In file included from /home/user/xpcc/src/xpcc/debug/logger.hpp:31:0,
                 from /home/user/xpcc/build/stm32f4_discovery/hard_fault/libxpcc/generated_platform/driver/core/cortex/hard_fault_handler.cpp:37:
/home/user/xpcc/src/xpcc/debug/logger/logger.hpp: In function 'void _hardFaultHandler(const uint32_t*, const uint32_t*)':
/home/user/xpcc/src/xpcc/debug/logger/logger.hpp:95:6: error: local frame unavailable (naked function?)
      *(xpcc::IOStream *) this << msg;
      ^
/home/user/xpcc/src/xpcc/debug/logger/logger.hpp:95: confused by earlier errors, bailing out
scons: *** [/home/user/xpcc/build/stm32f4_discovery/hard_fault/libxpcc/build/stm32f4_discovery/hard_fault/libxpcc/generated_platform/driver/core/cortex/hard_fault_handler.o] Error 1
scons: building terminated because of errors.
```

salkinium:
Hm, I've been misusing [`__attribute__((naked))`](https://gcc.gnu.org/onlinedocs/gcc/ARM-Function-Attributes.html#ARM-Function-Attributes):

> This attribute allows the compiler to construct the requisite function declaration, while allowing the body of the function to be assembly code. The specified function will not have prologue/epilogue sequences generated by the compiler. Only basic asm statements can safely be included in naked functions (see Basic Asm). While using extended asm or a mixture of basic asm and C code may appear to work, they cannot be depended upon to work reliably and are not supported. 

It's probably only sheer luck, that the startup script, which is also attributed `((naked))` compiles with `-O0`.
Let me fix that.

salkinium:
Please note that for `-O0` the stack size for the hard fault handler stack is probably too small.
It is currently set at 192 bytes, which has been measured with `-Os` only.

[...]

PR #113, opened 2016-03-19
Improve hardfault handler code

salkinium:
This change modifies the hardfault handler code to remove some bugs and add some features:
- Fix accidental masking of saving R0 in PSP, due to 8B stack alignment mask. This change uses LR to save R0 and saves SP into PSP.
- Remove breakpoint check for STM32F0, since it cannot access the `C_DEBUGEN` bit anyways.
- Use much simpler LED blinking code, which consumes less than 32B of stack.
- When using blinking LED only, switch stack only on stack overflow.
- Allow saving of high registers on ARM Cortex-M0 by moving them to the lower regs before pushing.
- Unifying the register printout for all ARM Cortex-M cores.
- Remove printing of stack, because it is useless information.
- Use better hardfault handler sizes in generated linkerscripts.
- Increase main stack size by ~1kB.

These changes are a great temporary solution until we have integrated [CrashCatcher](https://github.com/adamgreen/CrashCatcher).
This PR is related to #112.

Note: When using the blinking LED option only and no stack overflow occurred, the stack frames are preserved for GDB to unwind them. This means you get a blinking LED _and_ debuggable hardfault all in one!

cc @ekiwi @daniel-k @dergraaf

### DIFF
--- a/src/xpcc/architecture/platform/driver/core/cortex/driver.xml
+++ b/src/xpcc/architecture/platform/driver/core/cortex/driver.xml
@@ -7,7 +7,7 @@
 		</parameter>
 		<parameter name="enable_gpio" type="bool">true</parameter>
 		<parameter name="vector_table_in_ram" type="bool">false</parameter>
-		<parameter name="main_stack_size" type="int" min="512" max="8192">1856</parameter>
+		<parameter name="main_stack_size" type="int" min="512" max="8192">3040</parameter>
 
 
 		<parameter name="enable_hardfault_handler_log" type="enum" values="false;basic;true">false</parameter>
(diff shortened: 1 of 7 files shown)


salkinium:
@daniel-k @ekiwi: I've added blinking and debuggable hard fault code. So you can connect GDB and expect the stack frames to be intact.

Couldn't get it to work for the logging code, which pushes to the stack and confuses GDB.
Perhaps [manually adding unwind information](http://stackoverflow.com/a/33509137) might the way to go?
But the unwind tables aren't included anyways… Maybe its enough to have them in the `.elf` file.
Ideas?

salkinium:
Aaaaaaand I deleted the branch too quickly and let the CI fail. Again.
When will I learn this. _shakes head in dissapointment_

PR #115, opened 2016-03-30
Add TLSF allocator and enable by default

salkinium:
These changes add the [TLSF allocator](http://tlsf.baisoku.org) and enable it by default, completely replacing the newlib allocator.

Up to three arenas are constructed:
- SRAM: with `.heap1`, `.heap2` if available and `.heap3` if available
- CCM or DTCM: with `.heap0` if available
- ITCM: with `.heap4` if available

These three functions are provided:
- `malloc`: Uses SRAM by default, falls back to CCM if SRAM is full
- `realloc`: determines source arena by pointer address
- `free`: determines source arena by pointer address

Note: The fallback to CCM may not be desirable when the memory is used for DMA transfers, which cannot access the CCM.
At the moment nothing in xpcc can use DMA anyways, so I think that's a reasonable compromise.

cc @ekiwi @daniel-k

### DIFF
--- a/examples/stm32f469_discovery/tlsf-allocator/SConstruct
+++ b/examples/stm32f469_discovery/tlsf-allocator/SConstruct
@@ -1,0 +1,4 @@
+# path to the xpcc root directory
+xpccpath = '../../..'
+# execute the common SConstruct file
+execfile(xpccpath + '/scons/SConstruct')
--- a/examples/stm32f469_discovery/tlsf-allocator/main.cpp
+++ b/examples/stm32f469_discovery/tlsf-allocator/main.cpp
@@ -1,0 +1,54 @@
+#include <xpcc/architecture/platform.hpp>
+#include <tlsf.h>
+
+extern tlsf_t tlsf_heap0;
+extern uint32_t __heap1_start;
+
+int main()
+{
+	Board::initialize();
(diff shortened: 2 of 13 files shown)


ekiwi:
@salkinium I haven't looked at the source code yet, but one quick thought: The license might not be compatible, because - as far as I understand - there is no such thing as public domain in German law and thus the TLSF source code might be considered proprietary.

salkinium:
@ekiwi yes, exactly my concern too… :-(

ekiwi:
@salkinium could you try to contact the author and ask them if they could also release their allocator under BSD license?

salkinium:
@ekiwi I've contacted Matt, the author of this TLSF implementation, regarding relicensing.

There is the issue that TLSF consumes ~3kB of overhead per allocator. You could end up with up to ~9kB or dead RAM on systems with three RAM types. I think that's a heck of a lot.
We might have to modify this code to at least not have multiple 3kB overheads, but only one.

salkinium:
@mattconte has relicensed his TLSF implementation as BSD and pushed it to GitHub here, which is awesome:  https://github.com/mattconte/tlsf

I've adapted the PR to use this code now and modified the parameters so that the look up structures use less than 1kB.

PR #128, opened 2016-05-15
[ci] Test multiple ARM GCC versions.

strongly-typed:
Test with different compiler versions from https://launchpad.net/gcc-arm-embedded from gcc version 4.8.3 to 5.3.1.

Disable `-fdiagnostics-color=auto` again as this is the default and can be set by an environment variable locally on the user's host. But setting it will break compatibility with many gcc versions.

I should plant some trees to compensate for the additional CO2 emitted by TraivsCI now :-O


salkinium:
I'd be in favor of "swapping" the matrix, ie. grouping examples by compiler, instead of by board.
This way you can get a timely understanding of all the test results for _one_ compiler version (the newest), which I think is beneficial.

salkinium:
> Elapsed time 3 hrs 9 min 25 sec
> Total time 15 hrs 18 min 7 sec

Ähm…

strongly-typed:
> Elapsed time 3 hrs 9 min 25 sec

Yes, this is massive and not extremely efficient. That is why I need to plant some trees.
But I think it's worth the effort to brace oneself for unforeseen compiler glitches. (Remember the `std::round` and operator issues?).

> the newest

Which is **always** the best ;-)

With the reordered `.travis.yml` you get the results of `2016q1` first within about 20 minutes. That's not worse than before.

Actually the TravisCI jobs should be extended to check for different compilers on hosted (gcc-4.9 vs gcc-5 vs gcc-whatever and clang). And if possible we can use TravisCI's OS X environment.

salkinium:
> > Elapsed time 3 hrs 9 min 25 sec
> 
> Yes, this is massive and not extremely efficient. That is why I need to plant some trees.

Not so much concerned about the power consumption, more that a validation now takes over 3h. From <20mins to >3h, is ~10 times longer, which I'm not sure really benefits anyone.

I get why this is a good idea, and I support it in general, but Travis CI seems limits the builds to one at a time per repo (they'd probably get completely DoS'ed otherwise).
This means that if you have two PR going, it'll take 3h until the second one will be started.

I'm also not convinced we need to test all nine (9) compilers. Maybe the last four, or some selected ones, but not available versions.

This seems to me as more of a step you'd run on a release, rather than on every PR.
We want to catch many errors fast, which means not going to deep otherwise it takes forever.

I can imagine a test bot, which we tell to test this PR or a specific commit with your extended tests.
I will keep this PR open and investigate some solutions.

strongly-typed:
I fully acknowledge that this may overshoot a little. Did not expect that it takes sooooo much time.

> Maybe the last four

Not the last four, but selected ones. That's just enough if you look at two major distributions and want to include the latest version, too.

[Ubuntu Trusty 4.8.2](http://packages.ubuntu.com/trusty/devel/gcc-arm-none-eabi)
[Debian Jessie 4.8.4](https://packages.debian.org/stable/gcc-arm-none-eabi)
[Ubuntu Wily 4.9.3](http://packages.ubuntu.com/wily/gcc-arm-none-eabi)

> We want to catch many errors fast

Full ACK. So the sooner the committer knows that his or hers commit is incompatible with a compiler the sooner he or she can fix it. But I fully acknowledge that the TravisCI feedback for PRs should be as fast as possible (minutes, not hours).

> I can imagine a test bot

Own infrastructure? Maintenance? That's the benefit of TravisCI. It is just there at no cost.

Can we have a second `.travis.yml` (or a second account) with low priority jobs which does all the long running checks and a lightweight `.travis.yml` which does the 95% thing fast?

At the moment the library is generated about 100 times and the clock code is compiled many many times for the same target which is redundant and accumulates to long runtimes of the TravisCI job. That's also true for the UART and many other code which are included in many different examples. This is the result of so similar examples for different targets. Maybe it's worth to add more code to the `scons check=devices` [STM32 main.cpp](https://github.com/roboterclubaachen/xpcc/blob/develop/release/tests/stm32/main.cpp) which covers all modules (UART1, UART2, SPI1, ADC1, ...) once. That would be much much faster and the test coverage should be about the same.

salkinium:
> two major distributions and want to include the latest version, too.

👍 

> Own infrastructure? Maintenance? That's the benefit of TravisCI. It is just there at no cost.

Nah, there are bots out there that you can `@ping` and tell it to build the longer test.
We can use that before the PR is to be merged, and all the little things have been fixed.

> Can we have a second .travis.yml (or a second account) with low priority jobs which does all the long running checks and a lightweight .travis.yml which does the 95% thing fast?

A second Travis CI? [CircleCI?](https://circleci.com) Hehe.

> Maybe it's worth to add more code to the scons check=devices STM32 main.cpp which covers all modules (UART1, UART2, SPI1, ADC1, ...) once. That would be much much faster and the test coverage should be about the same.

Definitely!

salkinium:
Details of failures:
- `GCC 5.3.1 20160307` and `GCC 5.2.1 20151202`:

``` sh
Linking: /home/travis/build/roboterclubaachen/xpcc/build/lpcxpresso/lpc11c24/blink/blink.elf
collect2: fatal error: ld terminated with signal 11 [Segmentation fault], core dumped
compilation terminated.
```
- `GCC 4.8.4 20140725` and `GCC 4.8.4 20140526` in `examples/stm32f469_discovery/touchscreen`:

``` sh
main.cpp: In constructor 'LineDrawer::LineDrawer()':
main.cpp:15:35: error: invalid initialization of non-const reference of type 'xpcc::GraphicDisplay&' from an rvalue of type '<brace-enclosed initializer list>'
   c{Color::white(), Color::white()}
```

The latter is definitely fixable, the GNU linker segfault is much more tricky to solve.

TheTh0r:
I think the error in gcc 4.8.4 is correct, since reference in initialiser lists  are not allowed in c++11, but were allowed in later versions. If we stick to the c++11 standard we should change that. See http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2012/n3337.pdf, Section 8.5

salkinium:
> If we stick to the c++11 standard we should change that.

I agree. I need to have a look at the available compiler flags for GCC and enable this as an error for later versions.

salkinium:
Not so funny fact: Testing this PR generates 45793.35MB of cache data, since all the toolchains for all the combinations are cached.

salkinium:
I suspect something is badly broken with our cache settings, since other PRs take about 4-7GB of cache, and that's definitely too much. It used to be a few 100MBs.
Update: develop branch cache is 323.24MB large.

salkinium:
It's even worse than that: The timeout for downloading the cache is by default 3mins, but downloading a 7GB large cache and expanding it takes longer, so the cache download is aborted. Then the entire toolchain is built again for every test.
We could [increase the cache timeout](https://docs.travis-ci.com/user/caching/#Setting-the-timeout), but that would be fighting the symptom, not the cause.

strongly-typed:
I like to dismiss and close this PR now. Getting the latest ARM GCC is easy and well documented now. I might use CircleCI with different branches/tags of the Docker container to rebuild the functionality if it is really needed. I will just stick to 'latest' for now.

For AVR GCC there is TravisCI and CircleCI in place, testing the Ubuntu and the Atmel version in parallel.

Any objections?

salkinium:
> Any objections?

No, you may close it.

PR #129, opened 2016-05-15
[stm32] Fix I2C master reception of 2B transfers.

### LINKED ISSUE #127: STM32F103 I2C: When reading exactly two bytes the first byte is not acknowledged by the I2C master.
I stumbled over a coarse resolution of the `BMP180` sensor connected to an `STM32F103C8T6` at `I2cMaster1`. When reading exactly two bytes the first byte is not acknowledged by the master so that the slave will not send its second byte. The value is then always `0xff` and that resulted in a coarse temperature resolution.

Nevertheless the transaction returns `true`.

```
 this->transaction.configureWriteRead(buffer, 1, data.raw, 2);
```

screen shot 2016-05-15 at 11 55 16

When reading three or more bytes the first data read is acknowledged.

```
 this->transaction.configureWriteRead(buffer, 1, data.raw, 3);
```

screen shot 2016-05-15 at 11 53 58

With an `STM32F407` discovery board the two bytes read works correctly:
screen shot 2016-05-15 at 12 19 12

salkinium:
Fixes #127.

Credits go to @strongly-typed for finding the bug and fixing it!
cc @Sh4rK

### DIFF
--- a/src/xpcc/architecture/platform/driver/i2c/stm32/i2c_master.cpp.in
+++ b/src/xpcc/architecture/platform/driver/i2c/stm32/i2c_master.cpp.in
@@ -61,12 +61,13 @@
  *	- set start bit (RESTART!)
  *	#
  *	- check SB bit
- *	- set ACK low
+ *	- set ACK high
  *	- set POS high (must be used ONLY in two byte transfers, clear it afterwards!)
  *	- write address
  *	#
  *	- check ADDR bit
  *	- clear ADDR bit
+ *	- set ACK low
  *	#
  *	- check BTF bit
  *	- set STOP high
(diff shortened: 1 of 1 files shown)


niklas-arm:
@strongly-typed the reason why I don't immediately delete the branch is that Travis CI seems to build the PR one last time if it has been merged.
If the branch is deleted, Travis CI throws an error and we get a red cross, which I think is a little misleading.
```sh
The command "git fetch origin +refs/pull/129/merge:" failed and exited with 128 during .
```

salkinium:
@strongly-typed the reason why I don't immediately delete the branch is that Travis CI seems to build the PR one last time if it has been merged.
If the branch is deleted, Travis CI throws an error and we get a red cross, which I think is a little misleading.

``` sh
The command "git fetch origin +refs/pull/129/merge:" failed and exited with 128 during .
```

PR #131, opened 2016-05-16
Add isNotFull to queues to improve readability with more natural language.

strongly-typed:


salkinium:
I think you missed one:

``` sh
$ ack "isFull\(\)"
src/xpcc/architecture/platform/driver/i2c/stm32/i2c_master.cpp.in
632:    if (!queue.isFull() && transaction)
```

😛

salkinium:
Also, commit message length.

src/xpcc/architecture/driver/atomic/queue.hpp
```diff
@@ -67,6 +67,9 @@ namespace xpcc
 			ALWAYS_INLINE bool
 			isFull() const;
 			
+			ALWAYS_INLINE bool
+			isNotFull() const { return not isFull(); }
```
Sh4rK, line 71: I've noticed using `not` and other keyword versions of operators in xpcc, is this intentional? I haven't seen them used anywhere else in any C++ code, and I would think many people don't even know about these, so it might be confusing. On the other hand, I can accept that they can be more readable.

TheTh0r, line 71: I think Niklas introduced them and since he is the main developer of the code they start to prevail. I myself like the more common `!` more, since it is shorter and i am used to reading code with `!`, `&&` and `||`

salkinium, line 71: > I think Niklas introduced them and since he is the main developer of the code they start to prevail.

Yes, I introduced them intentionally to reduce confusion between logical and bitwise operators (`&&` vs. `&`, `||` vs. `|`, `!` vs `~`).

They are called Alternative Operators and are part of [the C++ language](http://en.cppreference.com/w/cpp/language/operator_alternative) (not part of C thought!).
We only use `and`, `or` and `not` though, and non of the other operators.
Yes, they are very rarely used even outside of embedded, and yes, most devs have never seen them 😛.

And prevail they should, as I think it significantly improves readability, especially for most logical "glue code" found in higher level drivers. I also find it tiring to visually parse the `!cond` for example, I find the `not` in `not cond` to be a much larger visual "target" to find.
It also makes me think of code that uses it as "more abstract" or "less to the metal", because I associate it with Python.

I'm always amazed at some of the push back I get for this though 😞.
As if this is the most "outrageous" feature of xpcc, when it is _already_ so different from the rest of the frameworks out there.

Ultimately, as Thorsten said, this is a matter of "being used to it", which I think is the main reason why coding conventions exist. I don't think there is any science behind "this is better to read", at least I can't find any significant studies on this.

Also, as Thorsten said, I'm one of the main developers, so it's my way or the highway.
Now excuse me, while I climb onto my high horse. Bwahaha.

TheTh0r, line 71: Yeah, I just stated my personal taste, I can agree to have these operators in new code. But probably it will be quite a long time until all `!`, `&&` and `||` are replaced.

salkinium, line 71: Yeah, I don't want to actively rewrite existing code, because that'll bloat xpcc's git history even more. So if I stumble across it, I'll change it.

The thing I'm most annoyed about is actually checked in trailing whitespace.
My editor strips trailing white space, so every time I modify such a file, I have to carefully stage all my changes. Grrrr.

daniel-k:
> It also makes me think of code that uses it as "more abstract" or "less to the metal", because I associate it with Python.

Yes, it feels like Python :smiley: And I really like it!

src/xpcc/architecture/driver/atomic/queue.hpp
Sh4rK, line 71: > I'm always amazed at some of the push back I get for this though

I didn't mean to offend you in any way, I was just genuinely curious about the use of these operators.

> Now excuse me, while I climb onto my high horse. Bwahaha.

Like [this](https://pocketsergeant.files.wordpress.com/2016/02/wp-1455927201540.jpg) one? :D

salkinium, line 71: > I didn't mean to offend you in any way, I was just genuinely curious about the use of these operators.

Oh, don't worry, I didn't read it that way.

I was thinking of some past discussions with other embedded developers. Very smart people, but somewhat… limited in their imagination. You know the "we've always done it this way"™ kind of mentality.

I like to 🔥 down the established conventions and build better ones in their place, but that's not always an option.

PR #135, opened 2016-05-17
[ui] Fix line drawing

Sh4rK:
With 5b8f924d947a9cc5d9299171768ac7863bd08b58 I changed how line drawing works, but as I figured out, this breaks at least `drawRectangle`, because it changes the length of lines drawn with `drawHorizontalLine` and `drawVerticalLine` (for length `n` they draw `n+1` pixel length lines, which is in a way correct, if the coordinate points are in the middle of the pixels, but at the same time seems odd).

This commit partly reverts that change, and partly modifies it, so that lines drawn with `drawLine` are still symmetric, but everything else works the old way.

### DIFF
--- a/src/xpcc/ui/display/buffered_graphic_display_impl.hpp
+++ b/src/xpcc/ui/display/buffered_graphic_display_impl.hpp
@@ -59,7 +59,7 @@ xpcc::BufferedGraphicDisplay<Width, Height>::drawHorizontalLine(
 	if (this->foregroundColor == glcd::Color::black())
 	{
 		const uint8_t mask = 1 << (start.getY() & 0x07);
-		for (uint_fast16_t x = start.getX(); x <= static_cast<uint16_t>(start.getX() + length); ++x) {
+		for (uint_fast16_t x = start.getX(); x < static_cast<uint16_t>(start.getX() + length); ++x) {
 			if( x < Width && y < Height ) {
 				this->buffer[x][y] |= mask;
 			}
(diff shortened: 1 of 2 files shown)


salkinium:
Hm, I have to admit that I pay too little attention to our GUI subsystem…

@daniel-k Were these rendering bugs not visible on the bot display?

daniel-k:
Well, there have been some minor gfx glichtes like this since the beginning IRC. Not sure if those were fixed, but if yes, these were high-level fixes in the widget system while this PR only touches low level primitives.

I honestly don't remember any rendering bugs on the display lately, but since it is mostly used during the competition under conditions where you don't care too much about this kind of problems, I might have just overlooked them.

Sh4rK:
You can see the issue by just drawing a single rectangle. On the right and bottom the lines will overshoot by 1 pixel.

salkinium:
:+1: tested and fix confirmed. Thanks!

salkinium:
Also as a side-note: Our graphics lib is quite horrible.

I spend 20mins figuring out why my white-on-black OLED display is not rendering horizontal lines in the rect that I was drawing to test this. [Then I found this "optimized" code](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/ui/display/buffered_graphic_display_impl.hpp#L59-L75).

(╯°□°)╯︵ ┻━┻

salkinium:
Commit manually cherry-picked on top of develop in https://github.com/roboterclubaachen/xpcc/commit/4b5f331f1a681621e40ce7a4f56633ccb5007fed.

Sh4rK:
> Also as a side-note: Our graphics lib is quite horrible.

Yes, there are strange things in it here and there :)
I actually have some "patches" (hacks) for it in my repo that are needed for my project, but not generally applicable (mostly there isn't even a correct way to do things, it seems colors were introduced, but not actually supported correctly)

salkinium:
> (mostly there isn't even a correct way to do things, it seems colors were introduced, but not actually supported correctly)

Yeah, it was written for very specific binary displays for AVR a long time ago, then "extended" for colors.

I have a partial NextGen implementation with [upainter](https://github.com/salkinium/upainter), with looooots of color, but I only have the lowest rendering layers and not much else. Kinda ran out of time now that I work.

PR #141, opened 2016-05-23
[scons] Add scons qtcreator alias

rleh:
This PR adds the _qtcreator_ alias from PR #71 to xpcc's global `SConstruct` file.


salkinium:
Thanks!

PR #146, opened 2016-06-19
Add startup hooks and linkerscript extensions

salkinium:
This adds two hooks to the startup script:
- `xpcc_hook_hardware_init()` to initialize hardware such as external SDRAM, and
- ~~`xpcc_hook_software_init()` to initialize the board before the main function.~~

The `hardware_init()` hook is called after the zero/copy of all internal memories and ARM Cortex-M initialization, but before zero/copy of all external memories, dynamic allocator init and libc_init.
It is therefore possible to place static data or bss regions into external SDRAM and have them initialize correctly.
~~The `software_init()` is called right before the main function and is primarily destined for initializing the board, so that the user does not have to.~~

The linkerscript can be extended using the parameter system, and arbitrary sections can be added for example in the `board.cfg`. See [this WIP for an example use case](https://github.com/salkinium/xpcc/commit/40897724562e0b1984bbe6019bc2e101def3b2d7#diff-68bd7e151b08f056fde5035d2f42bf72R11).

cc @ekiwi: See the addition of `string` as a parameter type, which includes `\t\n\r`. Are you ok with this?

### DIFF
--- a/src/xpcc/architecture/platform/driver/core/cortex/driver.xml
+++ b/src/xpcc/architecture/platform/driver/core/cortex/driver.xml
@@ -20,6 +20,10 @@
 		<parameter name="free_rtos_support" type="bool">false</parameter>
 		<parameter name="free_rtos_frequency" type="int" min="1000" max="10000">1000</parameter>
 
+		<parameter name="linkerscript_memory" type="string"></parameter>
+		<parameter name="linkerscript_sections" type="string"></parameter>
+		<parameter name="linkerscript_table_zero_extern" type="string"></parameter>
+		<parameter name="linkerscript_table_copy_extern" type="string"></parameter>
 
 		<!-- lpc -->
 		<template device-platform="lpc" out="linkerscript.ld">stm32/stm32_ram.ld.in</template>
--- a/src/xpcc/architecture/platform/driver/core/cortex/startup.c.in
+++ b/src/xpcc/architecture/platform/driver/core/cortex/startup.c.in
@@ -144,6 +144,10 @@ __libc_init_array(void);
 extern void
 exit(int) __attribute__ ((noreturn, weak));
 
+// default implementation is empty and does nothing
+void xpcc_hook_hardware_init(void) __attribute__(( weak ));
+void xpcc_hook_hardware_init(void) {}
+
 %% if parameters.allocator in ["block_allocator", "tlsf"]
 extern void __xpcc_initialize_memory(void);
 %% endif
@@ -274,6 +278,9 @@ Reset_Handler(void)
 	// Enable GPIO Clock
 	xpcc_gpio_enable();
 %% endif
+
+	// Initialize external memories in this hook
+	xpcc_hook_hardware_init();
 
 %% if parameters.allocator in ["block_allocator", "tlsf"]
 	// initialize xpcc block or TLSF allocator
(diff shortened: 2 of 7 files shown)


src/xpcc/architecture/platform/driver/core/cortex/driver.xml
```diff
@@ -20,6 +20,10 @@
 		<parameter name="free_rtos_support" type="bool">false</parameter>
 		<parameter name="free_rtos_frequency" type="int" min="1000" max="10000">1000</parameter>
 
+		<parameter name="linkerscript_memory" type="string"></parameter>
+		<parameter name="linkerscript_sections" type="string"></parameter>
+		<parameter name="linkerscript_table_zero_extern" type="string"></parameter>
+		<parameter name="linkerscript_table_copy_extern" type="string"></parameter>
```
ekiwi, line 26: Can you give me some examples of valid parameter values?

salkinium, line 26: From [WIP here](https://github.com/salkinium/xpcc/commit/90a94db07cbfa9ec682377d9597aff319ad35749#diff-68bd7e151b08f056fde5035d2f42bf72R11).

``` ini
core.cortex.0.linkerscript_memory =
    SDRAM (rwx) : ORIGIN = 0xC0000000, LENGTH = 16M

core.cortex.0.linkerscript_sections =
    .sdramdata :
    {
        __sdramdata_load = LOADADDR (.sdramdata);   /* address in FLASH */
        __sdramdata_start = .;                      /* address in RAM */

        KEEP(*(.sdramdata))

        . = ALIGN(4);
        __sdramdata_end = .;
    } >SDRAM AT >FLASH

    .heap_extern (NOLOAD) : ALIGN(4)
    {
        __heap_extern_start = .;
        . = ORIGIN(SDRAM) + LENGTH(SDRAM);
        __heap_extern_end = .;
    } >SDRAM

core.cortex.0.linkerscript_table_copy_extern =
    LONG (__sdramdata_load)
    LONG (__sdramdata_start)
    LONG (__sdramdata_end)
```

ekiwi:
> The software_init() is called right before the main function and is primarily destined for initializing the board, so that the user does not have to.

I understand the need for `xpcc_hook_hardware_init` as it is needed for setting up memories that need to be initialized before main is called. However, what is the use of `software_init()` when the next function to be executed is main anyways. Do you want to split up `xpcc` programs into `setup()` and `loop` functions? ;-)
I would prefer if we could keep the background magic to a minimum.

But maybe you have a good use case for `software_init()` that I am not seeing here.

src/xpcc/architecture/platform/driver/core/cortex/driver.xml
ekiwi, line 26: Hm. So `string` is just a wildcard type that could basically fit anything?
Personally I would rather prefer a more strictly typed approach, however, I do not have a complete solution at the moment.

salkinium:
> I would prefer if we could keep the background magic to a minimum.

Yeah, I wasn't too sure about this one. The only use case I found was calling `Board::initialize()` in the board targets, but it does seem a rather constructed use case.

src/xpcc/architecture/platform/driver/core/cortex/driver.xml
salkinium, line 26: Exactly, but I wasn't going to write a grammar for linker script language 😛.
Should I call this `rawstring` or `raw` instead?

salkinium:
Removed `xpcc_hook_software_init()`.

Issue #124, opened 2016-05-08
Register: error: use of deleted function


strongly-typed:
Trying to use `CtrlMeas::Mode1 | CtrlMeas::Mode0` with `register.hpp` in the plain blink example:

```
#include <xpcc/architecture/platform.hpp>
#include <xpcc/architecture/interface/register.hpp>

using namespace Board;

namespace xpcc
{

struct bme280
{
protected:

    enum class
    CtrlMeas : uint8_t
    {
        OSRS_T2 = xpcc::Bit7,
        OSRS_T1 = xpcc::Bit6,
        OSRS_T0 = xpcc::Bit5,
        OSRS_P2 = xpcc::Bit4,
        OSRS_P1 = xpcc::Bit3,
        OSRS_P0 = xpcc::Bit2,
        Mode1   = xpcc::Bit1,
        Mode0   = xpcc::Bit0,
    };
    XPCC_FLAGS8(CtrlMeas);

    enum class
    Mode : uint8_t
    {
        Sleep = 0,
        Forced = CtrlMeas::Mode0,
        Normal = CtrlMeas::Mode1 | CtrlMeas::Mode0,
    };

};

} // xpcc namespace

// ----------------------------------------------------------------------------
int
main()
{
    while (true)
        {}
    return 0;
}
```

With `gcc version 5.2.1 20151202`

```
arm-none-eabi-g++ -v
Using built-in specs.
COLLECT_GCC=arm-none-eabi-g++
COLLECT_LTO_WRAPPER=/opt/homebrew-cask/Caskroom/gcc-arm-embedded/5_2-2015q4,20151219/gcc-arm-none-eabi-5_2-2015q4/bin/../lib/gcc/arm-none-eabi/5.2.1/lto-wrapper
Target: arm-none-eabi
[...]

PR #151, opened 2016-06-28
[lisd3sdh] add missing casts for flags required by GCC 6.1

daniel-k:
With a recent `arm-none-eabi-gcc` this driver didn't compile anymore. Casting all the values seems to solve the problem.

### DIFF
--- a/src/xpcc/driver/inertial/lis3dsh.hpp
+++ b/src/xpcc/driver/inertial/lis3dsh.hpp
@@ -387,10 +387,10 @@ public:
 	FifoMode : uint8_t
 	{
 		Bypass = 0,
-		Fifo = FifoControl::FMODE0,
-		Stream = FifoControl::FMODE1,
+		Fifo = int(FifoControl::FMODE0),
+		Stream = int(FifoControl::FMODE1),
 		StreamTriggerFifo = int(FifoControl::FMODE1) | int(FifoControl::FMODE0),
-		BypassTriggerStream = FifoControl::FMODE2,
+		BypassTriggerStream = int(FifoControl::FMODE2),
 		BypassTriggerFifo = int(FifoControl::FMODE2) | int(FifoControl::FMODE1) | int(FifoControl::FMODE0),
 	};
 
@@ -398,14 +398,14 @@ public:
 	MeasurementRate : uint8_t
 	{
 		Off = 0,
-		Hz3_125 = Control4::ODR0,
-		Hz6_25 = Control4::ODR1,
+		Hz3_125 = int(Control4::ODR0),
+		Hz6_25 = int(Control4::ODR1),
 		Hz12_5 = int(Control4::ODR1) | int(Control4::ODR0),
-		Hz25 = Control4::ODR2,
+		Hz25 = int(Control4::ODR2),
 		Hz50 = int(Control4::ODR2) | int(Control4::ODR0),
 		Hz100 = int(Control4::ODR2) | int(Control4::ODR1),
 		Hz400 = int(Control4::ODR2) | int(Control4::ODR1) | int(Control4::ODR0),
-		Hz800 = Control4::ODR3,
+		Hz800 = int(Control4::ODR3),
 		Hz1600 = int(Control4::ODR3) | int(Control4::ODR0),
 	};
 
@@ -413,10 +413,10 @@ public:
 	Scale : uint8_t
 	{
 		G2 = 0,
-		G4 = Control5::FSCALE0,
-		G6 = Control5::FSCALE1,
+		G4 = int(Control5::FSCALE0),
+		G6 = int(Control5::FSCALE1),
 		G8 = int(Control5::FSCALE1) | int(Control5::FSCALE0),
-		G16 = Control5::FSCALE2,
+		G16 = int(Control5::FSCALE2),
 	};
 
 	enum class


salkinium:
:unamused: 
cc @strongly-typed and #124.

salkinium:
``` diff
- [lisd3sdh]
+ [lis3dsh]
```

The LIS3DSH is not the only driver that makes use of the register flags, so this is a bit of a larger problem.
We cannot upgrade to GCC 6.1 yet, since there is [no official precompiled version of it on launchpad](https://launchpad.net/gcc-arm-embedded) until 2016Q4.
So, this may not get merged so quickly, until I have a better solution for the "enum casting" issue.

ekiwi:
> We cannot upgrade to GCC 6.1 yet, since there is no official precompiled version of it on launchpad until 2016Q4.

What does this have to do with the availability of a new GCC version?
While in practice, we do trial and error programming with one specific compiler version, when in doubt, we should checkout whether this is a bug in the new GCC or (much more likely) if this is a bug in the older GCC version which did not correctly check the C++ standard compliance.

@salkinium as you are the author of the Flags stuff, what type is e.g. `FifoControl::FMODE0` supposed to be? Then we can check if this should implicitly convert to `uint8_t` or not.

salkinium:
> What does this have to do with the availability of a new GCC version?

It means I cannot reproduce the bug, cos I don't have GCC 6.1 and I'm not compiling it from source cos that's cray-cray. ;-)

> this is a bug in the older GCC version which did not correctly check the C++ standard compliance.

That's very likely. I think the fix is how it _should_ have been in the first place.

>  what type is e.g. `FifoControl::FMODE0` supposed to be?

This is a standard "enum class" of type `FifoControl`, which cannot be implicitly cast to anything. 

Note that this issue is independent of the `xpcc::Flags` mechanism, at least when only using the enum values (and not ORing them). 

> Then we can check if this should implicitly convert to `uint8_t` or not.

Nope. Implicitly converting the enums to an integer of any kind destroys the type safety of `xpcc::Flags`.
You cannot specify implicit casting operators for enum classes, since they are not "true" first-class types, but something special (total C++11 fail).

I think you will just always need to cast to `int` manually. 😞
What would be needed is to selectively allow the cast to int when assigning to other (unknown) enum class values, but that's not possible in C++.

ekiwi:
Thanks for your reply @salkinium.

As you point out the cast to int is "The right thing to do"(TM) anyway, so since this pull request does not break anything for your compiler version, but actually makes xpcc compile for Daniel's compiler version, can we just merge this?

salkinium:
> can we just merge this?

Hm. Please note, that using any other driver (say [`L3GD20`](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/driver/inertial/l3gd20.hpp#L223-L249)) will break this again.
It would be nice to have this fix as an atomic change over the entire code base, otherwise we are just creating unnecessary work for us in the future.

The changes are minimal, I can do them tonight. So please let's wait until then.

ekiwi:
> The changes are minimal, I can do them tonight. So please let's wait until then.

Great! Thank you.

I was only a little bit worried, that we would put off merging this until there is a prebuilt gcc6 for arm-none-eabi targets. I am sure that @daniel-k can test any changes that you make and comment on this pull request.

salkinium:
> I was only a little bit worried, that we would put off merging this until there is a prebuilt gcc6 for arm-none-eabi targets.

Nah, I was trying to stall and google up a better solution than casting 😛
But there doesn't seem one.

salkinium:
Ha, this is indeed an issue of not conforming to C++! I've solved this issue before for [our clang compiled unittests](https://github.com/roboterclubaachen/xpcc/commit/6591278818b176efae5d17b4a46f5bf0c9db7535). Since the drivers only get compiled with GCC, this never showed up before. Bad GCC!

salkinium:
`ack --ignore-dir=build "\t[A-Z][A-Za-z0-9]*?[ ]*?=.*?::"` shows the casts in question.

ekiwi:
Just an idea, without knowing the exact implementation of the Register class and macros:
Instead of using enum classes, couldn't we use a regular class with static members that have some custom type that implements relevant bit operations and can be implicitly casted to the underlying type, e.g. `uint8_t`?

salkinium:
> can be implicitly casted to the underlying type, e.g. uint8_t?

If it can be implicitly casted to the underlying type, you loose type safety. Then you can OR anything together, especially bits from different registers or register groups.
If you don't want that, you must not allow implicit casting, which brings you back to this problem, that you do actually want it to be implicitly cast, but only during "enum class value assignment time". We should write a proposal for C++20, bwahaha.

[I did explore the solution space in length before](http://blog.xpcc.io/2015/02/25/typesafe-register-access-in-c++/), especially also the "static constexpr" solution, which [I even prototyped](https://github.com/roboterclubaachen/xpcc/commit/ef55cb32a57b129af8a068f5b6c043eac2512312#diff-ba8846bac2db804c7b7c4a5d477002a0R159). The "static constexpr" prototype failed, because the linker requires the symbol to be defined in case it is referenced. This left only the "enum class" approach, which seem so to be the only one with compile time only definitions.

The only other sane approach I've seen is from [Biffle and his ETL framework](https://github.com/cbiffle/etl/blob/master/biffield/README.mkdn), but it targets on-device registers, not off-device registers like `xpcc::Flags`.

salkinium:
Closed in favor of #152.

salkinium:
Note that with C++17 [Inline Variables](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4424.pdf) will be supported that do not require an instantiation in a compilation unit. It might therefore be possible to implement this using static constexpr!

PR #153, opened 2016-06-30
[driver] Add double precision floating point calculation to BMP085.

strongly-typed:
And add a unittest that compares the fixed point results to the floating
point results.

### DIFF
--- a/release/tests/project.cfg
+++ b/release/tests/project.cfg
@@ -1,6 +1,3 @@
-[general]
-name = ${device}
-
 [build]
 device = $device
 clock = $clock
--- a/src/unittest_hosted.cfg
+++ b/src/unittest_hosted.cfg
@@ -1,5 +1,4 @@
 [general]
-name = unittest_hosted
 unittest = true
 
 [build]
--- a/src/xpcc/driver/pressure/bmp085.hpp
+++ b/src/xpcc/driver/pressure/bmp085.hpp
@@ -15,14 +15,19 @@
 #include <xpcc/architecture/interface/i2c_device.hpp>
 #include <xpcc/processing/timer.hpp>
 
+#include "bmp085_data.hpp"
+
 namespace xpcc
 {
 
-template < typename I2cMaster >
-class Bmp085;
-
 struct bmp085
 {
+
+using Calibration = xpcc::bmp085data::Calibration;
+using Data = xpcc::bmp085data::Data;
+using DataBase = xpcc::bmp085data::DataBase;
+using DataDouble = xpcc::bmp085data::DataDouble;
+
 protected:
 	/// @cond
 	/// The addresses of the Configuration and Data Registers
@@ -74,116 +79,6 @@ public:
 		UltraHighResolution = (0x03 << 6),
 	};
 
-	/**
-	 * Holds the calibration data from the sensor.
-	 * Values are used for calculation of calibrated
-	 * sensor values from raw sensor data
-	 */
-	struct ATTRIBUTE_PACKED
-	Calibration
-	{
-		int16_t  ac1;
-		int16_t  ac2;
-		int16_t  ac3;
-		uint16_t ac4;
-		uint16_t ac5;
-		uint16_t ac6;
-
-		int16_t  b1;
-		int16_t  b2;
-
-		int16_t  mb;
-		int16_t  mc;
-		int16_t  md;
-	};
-
-	struct ATTRIBUTE_PACKED
-	Data
-	{
-		template < typename I2cMaster >
-		friend class Bmp085;
-
-		// DATA ACCESS
-		inline Calibration &
-		getCalibration()
-		{
-			return calibration;
-		}
-
-		/**
-		 * Get the calibrated temperature for the device in 0.1 degree Celsius
-		 *
-		 * If recalculation is necessary it is done on the fly.
-		 * No I2C transaction.
-		 */
-		int16_t inline
-		getTemperature()
-		{
-			if (!(meta & TEMPERATURE_CALCULATED)) {
-				calculateCalibratedTemperature();
-			}
-			return calibratedTemperature;
-		}
-
-		/**
-		 * Get the calibrated pressure from the device in Pascal.
-		 *
-		 * If recalculation is necessary it is done on the fly.
-		 * No I2C transaction.
-		 */
-		int32_t inline
-		getPressure()
-		{
-			if (!(meta & PRESSURE_CALCULATED)) {
-				calculateCalibratedPressure();
-			}
-			return calibratedPressure;
-		}
-
-	private:
-		/**
-		 * Use the calibration data read from the sensor to
-		 * calculate the calibrated temperature from the
-		 * raw data.
-		 * The result is stored in this struct for further
-		 * access.
-		 */
-		void inline
-		calculateCalibratedTemperature();
-
-		/**
-		 * See calculateCalibratedTemperature()
-		 */
-		void inline
-		calculateCalibratedPressure();
-
-		// the order of these private variable is optimized for alignment of 4
-		Calibration calibration;
-
-		int16_t calibratedTemperature; // in 0.1 degree Celsius
-		int32_t calibratedPressure;    // in Pa
-
-		int32_t b5; // calculated in calculateCalibratedTemperature, needed for calculateCalibratedPressure
-
-		/// The raw data that was read from the sensor
-		/// 0 .. 1 temperature data
-		/// 2 .. 4 pressure data
-		uint8_t raw[5];
-
-		// bit 7-6: The mode in which the sensor operates
-		// bit 1: temperatureCalculated
-		// bit 0: pressureCalculated
-		uint8_t meta;
-
-		enum
-		{
-			/// Remember if the raw data was already converted to calibrated temperature
-			TEMPERATURE_CALCULATED = Bit1,
-			/// Remember if the raw data was already converted to calibrated pressure
-			PRESSURE_CALCULATED = Bit0,
-		};
-	};
-
 protected:
 	/// @cond
 	static constexpr uint8_t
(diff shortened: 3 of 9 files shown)


src/xpcc/driver/pressure/bmp085_data_impl_fp.hpp
```diff
@@ -0,0 +1,122 @@
...
+DataBase<CalculationMode::FixedPoint>::getTemperature(float &temp)
+{
+	temp = float(getTemperature() / 10.0f);
+}
+
+template <>
+void
+DataBase<CalculationMode::FixedPoint>::getTemperature(double &temp)
+{
+	temp = double(getTemperature() / 10.0);
```
salkinium: `10.0` is implicitly cast to float on our ARM targets to give us a chance to use the FPU instead of falling back to software routines for double operations.
It may be better to cast only `getTemperature()` then divide.

``` diff
-   temp = double(getTemperature() / 10.0);
+   temp = double(getTemperature()) / double(10.0);
```

src/xpcc/driver/pressure/bmp085_data.hpp
```diff
@@ -0,0 +1,191 @@
...
+	// Fixed Point
+	int16_t calibratedTemperature; // in 0.1 degree Celsius
+	int32_t calibratedPressure;    // in Pa
+
+	int32_t b5; // calculated in calculateCalibratedTemperature, needed for calculateCalibratedPressure
+
+private:
+	// Double Precision
+	// Prepared coefficients for temperature conversion in double precision
+	double c5, c6, mc, md;
```
salkinium, line 198: These 12 doubles are included for the fixed point as well.
That's 96 additional bytes that are not used for anything.
Please move this into the respective specializations.

src/xpcc/driver/pressure/bmp085_data.hpp
```diff
@@ -0,0 +1,191 @@
...
+// Forward declaration the test class
+class Bmp085Test;
+
+namespace xpcc
+{
+
+template < typename I2cMaster >
+class Bmp085;
+
+namespace bmp085data
```
salkinium, line 30: `bmp085`. The `data` is redundant and breaks with the established pattern for sensor data.

salkinium:
+9000 for unit tests!

src/xpcc/driver/pressure/bmp085_data.hpp
strongly-typed, line 30: There is already a `bmp085` struct.

src/xpcc/driver/pressure/bmp085.hpp
```diff
@@ -15,14 +15,20 @@
 #include <xpcc/architecture/interface/i2c_device.hpp>
 #include <xpcc/processing/timer.hpp>
 
+#include "bmp085_data.hpp"
+
+using namespace xpcc::bmp085data;
```
salkinium: This will polute the global namespace! 😱

src/xpcc/driver/pressure/bmp085_data.hpp
salkinium, line 30: Ah, I see. So `xpcc::bmp085data` is a `detail` namespace. Is the user ever expected to interact with it? No, right?

Your example uses `xpcc::bmp085data::Calibration`, perhaps you could alias that into the `bmp085` struct?

``` cpp
using Calibration = xpcc::bmp085data::Calibration;
```

src/xpcc/driver/pressure/bmp085.hpp
ekiwi: good catch!

src/xpcc/driver/pressure/bmp085_data.hpp
strongly-typed, line 198: Easily said ...

I use **function** template specialization to supply two different implementations for fixed point and double precision. I would need **class** template specialization to have a class for Fp (with member b5) and one for Dp (with members c5, ...).

The only solution I found so far is full class template specialization (as partial specialization is not allowed by the standard)

```
#include <stdio.h>

struct
FpData
{
    int m = 42;
};

struct
DpData
{
    double m = 4.2;
};

enum class
Mode
{
    Fix,
    Double,
};

template <Mode, class Storage>
class
Base
{
public:
    void
    calc();

private:
    Storage storage;
};

template<>
void
Base<Mode::Fix, FpData>::calc()
{
    printf("Calc Fix %d\n", storage.m);
}

template <>
void
Base<Mode::Double, DpData>::calc()
{
    printf("Calc Double %f\n", storage.m);
}

using Fp = Base<Mode::Fix, FpData>;
using Dp = Base<Mode::Double, DpData>;

int
main(void)
{
    Fp fp;
    Dp dp;

    fp.calc();
    dp.calc();

    return 0;
}
```

Any other idea?

salkinium, line 198: > Easily said ...

Yeah, I knew this was gonna be painful 😇

Can you try this?

``` cpp
template < CalculationMode >
class
DataBase
{
#ifdef __DOXYGEN__
// document the interface for this class
// it does not have to compile
#endif
};

/// @cond
// do not document the implementation!
template < >
class DataBase<FixedPoint>
{
    uint32_t bla;
// rest of actual implementation
};
// same for double class
/// @endcond
```

I've used a similar [method here](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/processing/resumable/resumable.hpp#L49-L96) to reduce the size of the result when specialization type is `void`.

salkinium, line 198: Also note, that your `getTemperature()` function can now have different return types!

strongly-typed, line 198: > Also note, that your `getTemperature()` function can now have different return types!

But why should it?

The aim of the encapsulation of the calculation is that the user code can be the same and only the driver differs in precision, speed and code size. I would add getters with references (`getTemperature(& int16_t)`, `getTemperature(& float)` and `getTemperature(& double)`)

BTW, I would suggest an xpcc::TemperatureFp data type which is `int16_t` in 0.01 degrees. That's between -320 (yeahh!!) and +320 degrees and is enough for all environmental sensors, including BME280.

salkinium, line 198: > But why should it?

Ok, I don't have a strong opinion on that. Having multiple getters is totally ok, and give the user more choice.

Regarding the fixed point class, [`modm::fixed_point_t`](https://github.com/salkinium/upainter/tree/master/ges/math) is the most feature complete, most tested and best integrating with normal code that I know of. Unfortunately it requires the relaxed constexpr rules of C++14 for correct compile time conversion and [xpcc still has C++11](https://github.com/roboterclubaachen/xpcc/issues/29).

(And of course you have to live with my shameless self plugging 😜)

strongly-typed, line 198: In your example the specialization fully overwrites the (in your case empty) class template, so that there can't be any common code or members (like `raw[5]` which is independent of FP/DP).

salkinium, line 198: > so that there can't be any common code or members (like raw[5] which is independent of FP/DP).

You could try to put common stuff into a base class and inherit the specializations from it.

Or: duplicate, duplicate, duplicate. Welcome to the painfree world of C++, baby! Bwahaha.

salkinium, line 198:
``` cpp
using xpcc::temperature_t = modm::fix16_t<7>;
```

2^-7 == 0.0078125 < 0.01, would leave signed 8-bit for integer part (-250 –> +250). No?

strongly-typed, line 198: You know how much I like duplication.

That's doing it quite well. Tried with `func()` in the base class but that did not work out. So that prototype must be repeated in every subclass.

```
#include <stdio.h>
#include <stdint.h>

enum class
Mode { Fix, Double };

template < Mode >
struct DataBase
{ uint8_t raw[5] = {1,2,3,4,5}; 
  // void func(); // Can't be here, must be in every subclass
  void baseFunc();
};

template< Mode m >
void DataBase<m>::baseFunc()
{ printf("base. "); }

struct Fp : public DataBase< Mode::Fix >
{ void func(); 
  int mem = 42; };

struct Dp : public DataBase< Mode::Double >
{ void func(); 
  double mem = 4.2; };

void Fp::func()
{ baseFunc(); printf("Fp: %d %d\n", raw[0], mem); }

void Dp::func()
{ baseFunc(); printf("Dp: %d %f\n", raw[1], mem); }

int
main(void)
{ Fp fp;     Dp dp;
  fp.func(); dp.func(); return 0;
}
```

salkinium, line 198: > That's doing it quite well. Tried with func() in the base class but that did not work out. So that prototype must be repeated in every subclass.

I was thinking that the base class is not a template and contains only the truly common things.
But yeah, duplication will not be completely gone.

strongly-typed, line 198: Yes, came to the same conclusion: No template necessary for the base class. And no `enum class CalculationMode`

salkinium:
Cool solution!
But doesn't the constructor `Bmp085(Data &data, uint8_t address=0x77);` now not need to take `DataBase &data` as the first argument?

src/unittest_hosted.cfg
```diff
@@ -1,5 +1,4 @@
 [general]
-name = unittest_hosted
```
salkinium, line 2: Please put in separate commit.
Same above.

release/tests/project.cfg
```diff
@@ -1,6 +1,3 @@
-[general]
-name = ${device}
-
```
salkinium, line 3: With "same above" I meant this change too. Please put it in the next commit.

strongly-typed:
Got it!

salkinium:
Cool, thanks for improving this and dealing with my review! Will merge once CI passes.

PR #155, opened 2016-07-02
Add F469I-DISCO display driver

salkinium:
This commits enables and configures the DSI and LTDC peripherals for 16-bit RGB565 color and adds a working `xpcc::GraphicDisplay` implementation.

cc @ekiwi @daniel-k

### DIFF
--- a/examples/stm32f469_discovery/display/SConstruct
+++ b/examples/stm32f469_discovery/display/SConstruct
@@ -1,0 +1,4 @@
+# path to the xpcc root directory
+xpccpath = '../../..'
+# execute the common SConstruct file
+execfile(xpccpath + '/scons/SConstruct')
--- a/examples/stm32f469_discovery/display/main.cpp
+++ b/examples/stm32f469_discovery/display/main.cpp
@@ -1,0 +1,39 @@
+#include <xpcc/architecture/platform.hpp>
+using namespace Board;
+using namespace xpcc::glcd;
+
+// ----------------------------------------------------------------------------
+int
+main()
+{
+	Board::initialize();
+	Board::initializeDisplay();
+
+	xpcc::GraphicDisplay& display = Board::getDisplay();
+	display.setFont(xpcc::font::Assertion);
+
+	display.drawRectangle(0,0, 10, 10);
+	display.drawRectangle(0,470, 10, 10);
+	display.drawRectangle(790,0, 10, 10);
+	display.drawRectangle(790,470, 10, 10);
+
+	display.drawRectangle(-1,-1, 10, 10);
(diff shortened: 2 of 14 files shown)


salkinium:
The display will now be reset on every call to `Board::initialize()`, but not initialized!
Instead on first call to `Board::getDisplay()` the display gets initialized and the `xpcc::GraphicDisplay` implementation gets constructed.
This does not turn on the display when you don't need it (like in the `blink` or `tlsf` examples).

ekiwi:
Very cool!

We don't have the Discovery boards to test this yet.
So if you want as to verify your pull request, you need to give us some time.

> Instead on first call to Board::getDisplay() the display gets initialized

A getter with lots of side effects... not sure if I like this.

Can't we hand a `enum class LCDDisplay {Enable, Disable}` to the `Board::initialize()` method or something like this?

salkinium:
> So if you want as to verify your pull request, you need to give us some time.

Don't worry, I included you and Daniel to show that xpcc can drive the display on the F469, in the hope that it'll influence your decision for the RCA display replacement 😜.

The initialization of the display is currently completely hardcoded with the exception of the color coding, which may also be ARGB8888, instead of the current RGB565.
However, xpcc's graphic stack has absolutely no hardware support or support for different color codings (it barely even qualifies for having any color support at all!), so I wanted to keep it very basic and compatible with what we currently have.
I believe adding support for the LTDC and DMA2D peripheral and integrating it into the graphics stack does not benefit xpcc or the RCA at the moment.

> Can't we hand a `enum class LCDDisplay {Enable, Disable}` to the `Board::initialize()` method or something like this?

I wouldn't modify `Board::initialize()` across boards at all. _frowns at `Uart::initialize(uint32_t)` on stm32_

The mechanism that we currently have is to add a `void initializeX()` method for each board "feature" (e.g. on [F3 Discovery](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/architecture/platform/board/stm32f3_discovery/stm32f3_discovery.hpp#L198-L243)). I would introduce a `void initializeDisplay()` which does the first half of the initialization, but still call it from `getDisplay()` on first call (which actually returns a singleton).

ekiwi:
we could always set a default value, if we add a enum parameter to the initialize method. This way if you want the default config, you just call `Board::initialize()` and if you want a specific config you call e.g. `Board::initialize(Board::Config::Lcd)`.
But I'm also fine with adding multiple separate initialize functions.

salkinium:
Now with touchscreen driver and drawing example. The FT6x06 TS driver performs quite well with 60Hz sample rate and no filtering required. I'm actually mildly impressed.

PR #158, opened 2016-07-02
[driver] Fix false assumption of ADC data in OSS != 0.

strongly-typed:
Not spotted by the unit test because test coverage is too small.


salkinium:
LÖL.

PR #168, opened 2016-08-14
Update all STM32 header and device files

salkinium:
Update to the latest CMSIS device headers for STM32 and adds a bunch of new devices.

cc @strongly-typed: You may need this for the new F7 devices.


salkinium:
> +566,152 −282,018

I wish there was a better way to deal with external dependencies like this. This change adds ~1MB to the entire repo. I want something like git submodules but automatically checked out.

I believe git subtree allows to store a squashed version of the files in the main repo. That may allow us to update ext files without increasing repo size. (I may be wrong).

strongly-typed:
```
 git clone --recursive
```

may be your friend. [Shallow submodules, too](http://stackoverflow.com/questions/2144406/git-shallow-submodules)

Is it worth a try?

salkinium:
> `git clone --recursive`

Argh, git is stupid here. It should do the `--recursive` automatically, if the repos `.gitconfig` has it enabled. I don't want to have to count on the user to have cloned xpcc with the right flags. That's simply not user-friendly.

AFAIK git subtrees "overlays" two repos in-place. So there is no additional step during cloning, it's just all one repo. Since we only rarely update `ext/` I would prefer this mechanism over submodules.

strongly-typed:
Could you please add the XML file for STM32F769NIH6, too? What is the correct device string in board.cfg?

salkinium:
> Could you please add the XML file for STM32F769NIH6, too?

I'll try. Adding completely new devices requires categorizing them first manually, so they can be merged with something.

> What is the correct device string in board.cfg?

`stm32f769ni` should do it. You may add the package `h6` too, but it's not being evaluated at the moment.

strongly-typed:
At the moment blinking, button and UART works magically with a [different processor](https://github.com/strongly-typed/xpcc/commit/2a25872735ba0da522bfa9b888fb05faefe7ca2f#diff-eca7d9dd4dd5946b5d69414dc5f3ffa9R2) on a STM32F769I Discovery Board. Aiming at double precision float support.

salkinium:
> Aiming at double precision float support.

Note, that it won't be supported out-of-the-box. You'll have to change the compiler flags in `arm.py`. There are more tweaks needed [to the startup code](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/architecture/platform/driver/core/cortex/startup.c.in#L249) and the build system will have to learn to [deal with the](https://github.com/roboterclubaachen/xpcc/blob/develop/scons/site_tools/xpcc.py#L469-L475) new [core `cortex-m7fd`](https://github.com/roboterclubaachen/xpcc/blob/develop/scons/site_tools/platform_tools.py#L378-L380).

ekiwi:
And you should carefully review our FreeRTOS port, in case you want to use it, as it contains routines to save and restore the floating point registers on context change.

salkinium:
I've added all new F7 device files. Note that I don't have hardware to test this on. You're the flauschiges Versuchskaninchen now.

src/xpcc/architecture/platform/devices/stm32/stm32f765_767_768_769_777_778_779-a_b_i_n_v_z-g_i.xml
```diff
@@ -0,0 +1,866 @@
...
+    <ram>610304</ram>
+    <core>cortex-m7f</core>
+    <pin-count device-pin-id="v">100</pin-count>
+    <pin-count device-pin-id="z">144</pin-count>
+    <pin-count device-pin-id="i">176</pin-count>
+    <pin-count device-pin-id="a">180</pin-count>
+    <pin-count device-pin-id="b">208</pin-count>
+    <pin-count device-pin-id="n">216</pin-count>
+    <header>stm32f7xx.h</header>
+    <define>STM32F4XX</define>
```
strongly-typed: Are you really sure about that? `<irony>`Maybe an else branch caught?`</irony>`
Same problem in `devices/stm32/stm32f745_746_756-b_i_n_v_z-e_g.xml`

salkinium: Used by our shitty FreeRTOS port, which requires STM32F4XX to be defined. So sort of semi-intentional.

salkinium: I've removed this from the device file now, if you want to use FreeRTOS on F7, please update the port to something not from the middle ages ;-P

salkinium:
Merged manually.

PR #174, opened 2016-08-24
Example: STM32F4 Discovery SPI

rleh:
There was a typo in the SPI example.
`SpiMaster2` has been initialized but `SpiMaster1` used in the while loop.

### DIFF
--- a/examples/stm32f4_discovery/spi/main.cpp
+++ b/examples/stm32f4_discovery/spi/main.cpp
@@ -16,7 +16,7 @@ main()
 
 	while (1)
 	{
-		SpiMaster1::transferBlocking(0xF0);
+		SpiMaster2::transferBlocking(0xF0);
 	}
 
 	return 0;
--- a/examples/stm32f4_discovery/spi/main.cpp
+++ b/examples/stm32f4_discovery/spi/main.cpp
@@ -1,5 +1,4 @@
 #include <xpcc/architecture/platform.hpp>
-#include <xpcc/architecture/platform.hpp>
 
 
 int
--- a/examples/stm32f4_discovery/spi/main.cpp
+++ b/examples/stm32f4_discovery/spi/main.cpp
@@ -1,21 +1,23 @@
 #include <xpcc/architecture/platform.hpp>
 
+// Use SPI 2
+using SpiMaster = SpiMaster2;
 
 int
 main()
 {
 	Board::initialize();
 
-	// Enable SPI 2
-	GpioOutputB12::connect(SpiMaster2::Nss);
-	GpioOutputB15::connect(SpiMaster2::Mosi);
-	GpioInputB14::connect(SpiMaster2::Miso);
-	GpioOutputB13::connect(SpiMaster2::Sck);
-	SpiMaster2::initialize<Board::systemClock, MHz20>();
+	// Enable SPI
+	GpioOutputB12::connect(SpiMaster::Nss);
+	GpioOutputB15::connect(SpiMaster::Mosi);
+	GpioInputB14::connect(SpiMaster::Miso);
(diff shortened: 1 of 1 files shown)


strongly-typed:
[It broke two years ago :-(](https://github.com/roboterclubaachen/xpcc/commit/96df5bbb917672b35c15faeea308ac73994c47ce?diff=unified#diff-2a62cdf456b058d94c1d9e4311fb9b88)

salkinium:
Could you make the `SpiMaster2` a typedef, so this can never happen again?

ekiwi:
> Could you make the SpiMaster2 a typedef, so this can never happen again?

I'd like to vote against that in order to keep the magic (i.e. indirection) to a minimum in the examples.

salkinium:
> magic (i.e. indirection) to a minimum

A `typedef` is not magic, it is best practice. We'd always `typedef xpcc::SoftwareSpiMaster<Sck, Mosi, Miso> SensorSpiMaster` and not copy it around everywhere. Same applies for a hardware SPI masters.

ekiwi:
> Same applies for a hardware SPI masters.

I disagree:
A hardware SPI has a fixed meaning and is easily cross referenced with the data sheet. The manufacturer has given it a name and we should stick with that name in order to avoid additional complexity.
A software SPI on the other hand is the users "creation" and thus they are responsible for naming it in a sensible manner.

salkinium:
This example is so pointless (but thanks for making it less broken ;-).

We're connecting `GpioOutputB12::connect(SpiMaster2::Nss);`, even though we can never use it, because we don't have a `SpiSlave`. In fact, it's probably a very bad idea to connect Nss to a hardware SPI, because if that pin goes low, the peripheral switches to slave mode, and our driver can't deal with it.

And then we're sending `0xF0`, but not printing what we read? This can't even be used as a loop-back test without modification.

What is this example supposed to show exactly? That our documentation isn't lying? Shouldn't this rather be part of the documentation of the SpiMaster interface?

It's not even specific to the STM32F4, certainly not to the SPI2 of the STM32F4. This could (and should) work on the AVR, or the STM32F1 or the F3.

> The manufacturer has given it a name and we should stick with that name in order to avoid additional complexity.

We'd be advocating hardcoding the instance of the SPI driver. How is that useful?

The `typedef` would be a good opportunity to explain that all SPI instances share the same interface, even the bit-banged one, but that you must change the Gpio connections and perhaps the baudrate, otherwise you'd get a compile error.
That simple comment would already make this example like 9001% more useful.

I think it's obvious enough that `SpiMaster2` refers to `SPI2` on all the STM32, but it may not be obvious that you can use the `SoftwareSpiMaster` as a drop-in replacement.

I'd even go a step further and advocate that you should ALWAYS use the bit-banged masters for SPI and I2C especially when developing an external chip driver, since they are more stable than the hardware drivers (see #127). You can then switch to the hardware implementations and see if it still works.

strongly-typed:
ALWAYS is ALWAYS wrong ;-)

> I'd even go a step further and advocate that you should ALWAYS use the bit-banged masters for SPI [...]

Well, that's what I thought until last year. I discovered the known _anomality_ that the ENC28J60 [to run the SPI at frequencies of at least 8 MHz.](http://ww1.microchip.com/downloads/en/DeviceDoc/80257d.pdf) (at least ist Rev. B4 Silicon).

salkinium:
> I discovered the known anomality that the ENC28J60 to run the SPI at frequencies of at least 8 MHz. (at least ist Rev. B4 Silicon).

That's it. I'm becoming a farmer.

salkinium:
Merged and modified manually in https://github.com/roboterclubaachen/xpcc/commit/b3278777868a4549863ddbefc1d9e31484e7d8f2.

PR #179, opened 2016-09-02
[wip] Add a porting guide

salkinium:
We've had a few requests for porting xpcc to new platforms, and this is how you do it.

Note, that this is work in progress, and since I haven't ported any platform yet, it's pretty theorical until someone tries it out.

If you are the one trying this out and you're stuck, or you've found that the guide is missing information, please leave a comment.
For bigger things, write us an email.

[Rendered guide](https://github.com/salkinium/xpcc/blob/feature/porting_guide/PORTING.md).

### DIFF
--- a/PORTING.md
+++ b/PORTING.md
@@ -1,0 +1,385 @@
+# Porting to xpcc
+
+## What xpcc already supports
+
+xpcc contains extensive support for AVR and STM32 devices, specifically:
+
+- AT90, ATtiny, ATmega devices
+- STM32 F0, F1, F3, F4, F7 devices.
+
+If your device falls into any of those categories, you can just use the specific device
+identifier in your `project.cfg` and it'll very likely just work out-of-the-box.
+
+If you want to add a development board for an existing device, have a look at [`xpcc/architecture/platform/board`](https://github.com/roboterclubaachen/xpcc/tree/develop/src/xpcc/architecture/platform/board).
+
+## Adding support for new AVR or STM32 devices
+
+It may be as easy as adding a new device file, but if it's a brand new device, [the device headers also need to be updated](https://github.com/roboterclubaachen/xpcc/tree/develop/ext/). When porting a new series (for example the STM32L4), a lot of existing code can be reused, but there may be very important differences between the two that are not immediately obvious to someone who isn't that familiar with STM32 devices.
+
+We recommend [writing us an email][mailing_list], or [opening up an issue on GitHub](https://github.com/roboterclubaachen/xpcc/issues).
+
+## Porting a new device from a new vendor
+
+Note: We assume this new device is Cortex-M based. For other CPU architectures, [send us an email][mailing_list].
+
+Porting to a new vendor device requires you to:
+
+1. Add a bunch of header files for the new device.
+2. Add a device files for the new device.
+3. Make the build system aware of the new device classes.
+4. Adapt the Cortex-M startup- and linkerscripts.
+5. Compile a minimal example.
+6. Write a basic GPIO and delay driver.
+7. Compile an LED blinking example.
+
+We'll discuss every point in detail for the rest of this chapter.
+
+### Adding device header files
+
+Copy the [CMSIS header files](https://www.arm.com/products/processors/cortex-m/cortex-microcontroller-software-interface-standard.php)  for the device family into the `ext/` folder. You can usually download them as part of a driver pack from the vendor's website.
+Have a look at [the STM32 device headers](https://github.com/roboterclubaachen/xpcc/tree/develop/ext/st) as an example for what you are looking for.
+
+Create a `README.md` that contains information on the source of the header files, with a link to the vendor website, the download date/version and most importantly the license text.
+Don't forget to add the include path of the device files to the [`ext/SConstruct`](https://github.com/roboterclubaachen/xpcc/blob/develop/ext/SConscript) file!
+
+
+### Adding device files
+
+The [xpcc device files](https://github.com/roboterclubaachen/xpcc/tree/develop/src/xpcc/architecture/platform/devices) contain the information that is used to generate the HAL. For AVR and STM32 devices these files are generated from another (vendor specific) data source.
+However, for a new vendor such machine readable data might not be available and if it is, it will require some work to create the necessary parser to automatically generate device files.
+We recommend building a minimal proof-of-concept port for one device first.
+Once you have one device working, you can get started working on an automatic parser if you want to support a wide range of devices.
+
+You need to manually create a new device file for your device. We recommend copying [this device file for the STM32F303](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/architecture/platform/devices/stm32/stm32f303-c_k_r-6_8.xml) to a new folder named after the platform (like `devices/sam`).
+
+Vendors usually have a naming convention of their devices, make sure the device attributes and file name reflect this. The semantics of the vendor's device naming scheme can sometimes be gathered directly from the datasheet or by comparing different devices (i.e. their names and capabilities).
+For example, the `SAM4LC4B` identifier can be split like this:
+```xml
+  <device platform="sam" family="4" name="lc" pin_id="b" size_id="4">
+    <!-- ... -->
+  </device>
+```
+
+Modify the copied device file to reflect the new devices data.
+Since you only have one specific device to port, you can throw out the duplicate XML tags which are normally filtered by `device-` attributes.
+
+Adding the total FLASH and RAM sizes allows the build system to display accurate static resource usage and to warn you if your program does not fit into memory:
+```xml
+<flash>65536</flash>
+<ram>16384</ram>
+```
+
+Choose the correct ARM Cortex-M core (`m0`, `m3`, `m4`, `m4f`, `m7`, `m7f`, `m7fd`):
+```xml
+<core>cortex-m4f</core>
+```
+
+The CMSIS device headers you just added are now available in the include path, so you should add the device header:
+```xml
+<header>sam4.h</header>
+```
+
+Often you must not include the device specific device header directly, but let the family device header choose it for you based on a preprocessor define, which you can provide here too:
+```xml
+<define>SAM4LC4B</define>
+```
+
+You need to include at least three drivers: `core`, `clock` and `gpio`.
+
+Make sure you give the `core` driver the right information about the memory map. This information is used to generate the linkerscript.
+Don't worry about different memories right now, just add the main FLASH and one SRAM region here (called `sram1`). The `size` unit is kB.
+```xml
+<driver type="core" name="cortex">
+  <memory access="rx" start="0x8000000" name="flash" size="64"/>
+  <memory access="rwx" start="0x20000000" name="sram1" size="12"/>
+```
+
+Add all or part of the vector table. No need to add the Cortex-M exceptions, we do that for you.
+The `name` will be appended with `_Handler`, so declaring `extern "C" void WWDG_Handler(void) {...}` anywhere will link that function into the vector table (note the C linkage!).
+```xml
+  <vector position="0" name="WWDG"/>
+  <vector position="1" name="PVD"/>
+  <!-- ... -->
+  <vector position="81" name="FPU"/>
+</driver>
+```
+
+The `clock` driver just needs to be included, we'll hardcode the specifics in code:
+```xml
+<driver type="clock" name="sam4"/>
+```
+
+The `gpio` driver should contain some GPIOs, just add the ones you need. Adding more later is cheap. Ignore the alternate functions for now, you only want to blink an LED.
+```xml
+<driver type="gpio" name="sam4">
+  <gpio port="A" id="0"/>
+  <gpio port="A" id="1"/>
+<driver>
+```
+
+Remove all other drivers.
+
+### Making the build system aware
+
+Our build system is organically grown, er, I mean, *has* organically grown. Therefore it is a bit hacky in places and you now need to make it aware of the new devices from a new vendor:
+
+1. [Split the device identifier into the vendor naming scheme useful](https://github.com/roboterclubaachen/xpcc/blob/7b32f74cc43c7bfaf815287902ef2bd4a6baa84d/tools/device_files/device_identifier.py?ts=4#L86-L95).
+2. [Find the device file using the naming scheme](https://github.com/roboterclubaachen/xpcc/blob/3c7cd31e5aad66e47d073dc445551a21416d21eb/scons/site_tools/platform_tools.py?ts=4#L92-L103).
+3. [Thou Shalt ~~Not~~ Pass](https://github.com/roboterclubaachen/xpcc/blob/7b32f74cc43c7bfaf815287902ef2bd4a6baa84d/tools/device_files/device.py#L134-L139).
+4. [Add some platform-specific tests for the new devices](https://github.com/roboterclubaachen/xpcc/blob/3c7cd31e5aad66e47d073dc445551a21416d21eb/scons/site_tools/platform_tools.py?ts=4#L385-L414). These tests are [custom Jinja2 filters](http://jinja.pocoo.org/docs/dev/api/#custom-filters) that can conveniently be used in the driver templates.
+
+### Adapt the startup- and linkerscripts
+
+The [Cortex-M startup script](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/architecture/platform/driver/core/cortex/startup.c.in?ts=4) is pretty universal. These are the steps it does:
+
+1. Generate the exception table and places it at the beginning of Flash.
+2. Fill the stack with magic values to catch unintentional stack execution.
+3. Call platform specific code (enabling internal memories, **needs porting**).
+4. Copy and zero internal memories.
+5. Initialize the Cortex-M CPU.
+6. Call `xpcc_gpio_enable()` to enable clock to GPIO (**needs porting**).
+7. Call external hardware initialize hook.
+8. Copy and zero external memories.
+9. Initialize heap on all memories.
+10. Call libc init incl. execution of C++ static constructors.
+11. Call `main()`.
+
+We'll write the `xpcc_gpio_enable()` function when we port the GPIO driver.
+
+To provide code for step 3, copy and adapt [the macro file for the STM32 devices](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/architecture/platform/driver/core/cortex/stm32/stm32.macros?ts=4).
+You can leave the `defines()` and `appendInterrupt()` macros as they are, but you need to implement the `startupCode()` macro with whatever platform specific hardware needs to be initialized. Don't do complicated stuff here, the system is not fully booted yet!
+
+The [simplest linkerscript available is `stm32/stm32_ram.ld.in`](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/architecture/platform/driver/core/cortex/stm32/stm32_ram.ld.in?ts=4), which fits everything into one SRAM region.
+The linkerscipt diagram in the comment gives you an overview of what goes where, check that against the devices memory map. Note that while the diagram shows specific addresses, the section addresses and sizes are taken from the device file!
+
+You need to tell the [core driver which linkerscript to use for you device](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/architecture/platform/driver/core/cortex/driver.xml?ts=4#L29-L40):
+```xml
+<template device-platform="sam" out="linkerscript.ld">stm32/stm32_ram.ld.in</template>
+```
+
+### Compile a minimal example
+
+At this point you should test your changes to the build system and the port of the core driver.
+Create an minimal example in `xpcc/examples/sam4/minimal`, containing:
+
+- the default `SConstruct` file:
+```python
+# path to the xpcc root directory
+xpccpath = '../../..'
+# execute the common SConstruct file
+execfile(xpccpath + '/scons/SConstruct')
+```
+
+- a minimal `main.cpp` file:
+```cpp
+/* Copyright (c) {{ year }}, {{ your name }}
+ * All Rights Reserved.
+ *
+ * The file is part of the xpcc library and is released under the 3-clause BSD
+ * license. See the file `LICENSE` for the full license governing this code.
+ */
+#include <xpcc/architecture/platform.hpp>
+
+// Leave this function empty for now
+extern "C" void xpcc_gpio_enable(void) {}
+
+int main()
+{
+    // Initialize one GPIO output manually using the datasheet.
+    // Don't forget to enable the GPIO clock!
+    // GPIO_PORTA_OUTPUT = (1 << 0);
+
+    while(1)
+    {
+        // Toggle the pin manually
+        // GPIO_PORTA_TOGGLE = (1 << 0);
+
+        // Create a very crude busy wait. The numbers don't have to
+        // be super accurate, just in the ball-park of a 1s delay.
+        // Use the initial CPU frequency in MHz here.
+        constexpr uint32_t fcpu_MHz = 8;
+        for (volatile uint32_t ii = 0; ii < fcpu_MHz * 250000; ii++) ;
+    }
+}
+```
+
+- the `project.cfg` file:
+```ini
+[build]
+device = sam4lc4b
+buildpath = ${xpccpath}/build/sam4/${name}
+```
+
+To compile this example, `cd` into the folder and execute `scons`.
+If everything goes well, there will be warnings about the missing `clock` and `gpio` drivers, but at the end there should be a printout of the resource usage, similar to this (but with realistic numbers):
+```sh
+cd examples/sam4/minimal
+scons
+[...]
+Memory Usage
+------------
+Device: sam4lc4b
+
+Program:    1000 bytes (1.0% used)
+(.data + .reset + .rodata + .text)
+
+Data:       1000 bytes (2.0% used) = 1 bytes static (1.0%) + 1000 bytes stack (1.0%)
+(.bss + .data + .noinit + .stack)
+
+Heap:      9999 bytes (97.0% available)
+(.heap1)
+```
+
+You can explore the generated code for the core driver in `xpcc/build/sam4/minimal/libxpcc/generated_platform/driver/core/cortex/`.
+
+If your development board is natively supported via OpenOCD (any of the scripts inside `openocd/share/openocd/scripts/`), you can just add it to your `project.cfg` file:
+```ini
+[openocd]
+configfile = board/atmel_sam4l8_xplained_pro.cfg
+```
+For custom setups, [read this](http://xpcc.io/reference/build-system/#project-configuration).
+
+You should now be able to program the device using `scons program`.
+Check if the LED is blinking at a constant interval.
+
+If the LED is not blinking and you have OpenOCD connected, you can [launch a GDB debug session](https://github.com/roboterclubaachen/xpcc/tree/develop/examples/stm32f3_discovery/gdb).
+
+If something goes wrong in the building step, check the previous steps, maybe you missed something, or we forgot to add something to this guide.
+If you're totally stuck [write us an email][mailing_list].
+
+### Porting the delay functions
+
+To keep it simple, we'll leave the device running on the clock it boots up on. This is usually some internal oscillator in the MHz range.
+In order to enable the accurate busy-wait delay functions (`xpcc::delay{Nano|Micro|Milli}seconds(uint16_t)`) you need to provide some information about the frequency that the CPU core is running at.
+
+- `fcpu` contains the CPU frequency in Hz
+- `fcpu_kHz` contains the CPU frequency in kHz
+- `fcpu_MHz` contains the CPU frequency in MHz
+- `ns_per_loop` contains the ns per busy-wait delay loop (processor dependent)
+
+Create two files in the `xpcc/architecture/platform/driver/clock/sam4` folder:
+
+- `clock.hpp` (see [STM32 version for verbatim](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/architecture/platform/driver/clock/stm32/clock.hpp.in?ts=4)):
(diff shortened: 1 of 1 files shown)


PORTING.md
```diff
@@ -0,0 +1,327 @@
...
+
+### Making the build system aware
+
+Our build system is organically grown, er, I mean, *has* organically grown. Therefore it is a bit hacky in places and you now need to make it aware of the new devices from a new vendor:
+
+1. [Split the device identifier into the vendor naming scheme useful](https://github.com/roboterclubaachen/xpcc/blob/7b32f74cc43c7bfaf815287902ef2bd4a6baa84d/tools/device_files/device_identifier.py?ts=4#L86-L95).
+2. [Find the device file using the naming scheme](https://github.com/roboterclubaachen/xpcc/blob/3c7cd31e5aad66e47d073dc445551a21416d21eb/scons/site_tools/platform_tools.py?ts=4#L92-L103).
+3. [Thou Shalt ~~Not~~ Pass](https://github.com/roboterclubaachen/xpcc/blob/7b32f74cc43c7bfaf815287902ef2bd4a6baa84d/tools/device_files/device.py#L134-L139).
+4. [Add some platform tests for the new devices](https://github.com/roboterclubaachen/xpcc/blob/3c7cd31e5aad66e47d073dc445551a21416d21eb/scons/site_tools/platform_tools.py?ts=4#L385-L414).
+
```
ekiwi, line 130: Maybe add something like this:

``` .md
These are some convenient [jinja2 filters](http://jinja.pocoo.org/docs/dev/api/#custom-filters) that can later be used in the driver templates.
```

PORTING.md
```diff
@@ -0,0 +1,327 @@
...
+### Adapt the startup- and linkerscripts
+
+The [Cortex-M startup script](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/architecture/platform/driver/core/cortex/startup.c.in?ts=4) is pretty universal. These are the steps it does:
+
+1. Generate the exception table and places it at the beginning of Flash.
+2. Fill the stack with magic values to catch unintentional stack execution.
+3. Call platform specific code (enabling internal memories, **needs porting**).
```
ekiwi, line 137: This is one of the ugly parts left over from when I first tried to unify the startup code, however as long as it is still used we should point people to [where the platform definitions are included](https://github.com/roboterclubaachen/xpcc/blob/8266db98997f27c68880a48818425b143e7e1611/src/xpcc/architecture/platform/driver/core/cortex/startup.c.in#L45)

Or maybe just inline the code from [stm32.macros](https://github.com/roboterclubaachen/xpcc/blob/8266db98997f27c68880a48818425b143e7e1611/src/xpcc/architecture/platform/driver/core/cortex/stm32/stm32.macros) and [lpc.macros](https://github.com/roboterclubaachen/xpcc/blob/8266db98997f27c68880a48818425b143e7e1611/src/xpcc/architecture/platform/driver/core/cortex/lpc/lpc.macros)?

PORTING.md
```diff
@@ -0,0 +1,327 @@
...
+4. Copy and zero internal memories.
+5. Initialize the Cortex-M CPU.
+6. Call `xpcc_gpio_enable()` to enable clock to GPIO (**needs porting**).
```
ekiwi, line 140: > In `xpcc` GPIOs are assumed to be usable without explicitly initializing the subsystem.
> For other peripheral IP, the clock is enabled in the `initialize` method that needs to be explicitly called by the user.

PORTING.md
```diff
@@ -0,0 +1,327 @@
...
+7. Call external hardware initialize hook.
```
ekiwi, line 141: Haha, this part is a nice example of the comments and the code disagreeing (or at least being somewhat ambiguous).

[from startup.c.in](https://github.com/roboterclubaachen/xpcc/blob/8266db98997f27c68880a48818425b143e7e1611/src/xpcc/architecture/platform/driver/core/cortex/startup.c.in#L293)

``` .c
    // Initialize external memories in this hook
    xpcc_hook_hardware_init();
```

PORTING.md
```diff
@@ -0,0 +1,327 @@
...
+8. Copy and zero external memories.
+9. Initialize heap on all memories.
+10. Call libc init incl. execution of C++ static constructors.
+11. Call `main()`.
+
+We'll write the `xpcc_gpio_enable()` function when we port the GPIO driver.
+
+To provide code for step 3, copy and adapt [the macro file for the STM32 devices](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/architecture/platform/driver/core/cortex/stm32/stm32.macros?ts=4).
+You can leave the `defines()` and `appendInterrupt()` macros as they are, but you need to implement the `startupCode()` macro with whatever platform specific hardware needs to be initialized. Don't do complicated stuff here, the system is not fully booted yet!
```
ekiwi: ah sorry, I did not see this. Please ignore the comment above.

PORTING.md
```diff
@@ -0,0 +1,327 @@
...
+(.heap1)
+```
+
+You can explore the generated code for the core driver in `xpcc/build/sam4/empty/libxpcc/generated_platform/driver/core/cortex/`.
+
+We'll defer flashing onto the device, until we can actually blink an LED.
+Without a debugger you won't be able to tell what is going.
+
+If something goes wrong in the building step, check the previous steps, maybe you missed something, or we forgot to add something to this guide.
+If you're totally stuck write us an email (try to work the problem first though ;-).
```
ekiwi: most controversial part of this pull request: close with one or two brackets? ( ;-) )

PORTING.md
```diff
@@ -0,0 +1,327 @@
...
+
+### Porting the clock driver
```
ekiwi: A more natural order would imho be to first port the GPIO driver.
It is much more straight forward compared to the clock driver and it actually provides you with a visible result.

ekiwi: A compromise could be to tell people to test their code first by blinking an LED through direct writes to registers.

PORTING.md
```diff
@@ -0,0 +1,327 @@
...
+
+To keep it simple, we'll leave the device running on the speed it boot up on. This is usually some internal oscillator in the MHz range.
```
ekiwi: I really need to keep on reading and not review this top down.
But how about changing the name of the section to e.g. `Including initial clock information` (maybe you can think of a better way to phrase this...)

PORTING.md
```diff
@@ -0,0 +1,327 @@
...
+
+}
+}
+```
+
+### Porting the GPIO driver
+
+The GPIO driver is the most important driver of them all, because it gives you bit-banged versions of SPI and I2C throught the `xpcc::SoftwareSpiMaster` and `xpcc::SoftwareI2cMaster` classes.
+
+The [STM32 GPIO driver is quite complicated](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/architecture/platform/driver/gpio/stm32/gpio.hpp.in?ts=4), because we're a little weird (lol), but the only methods need to implement right now are [these ones from the GPIO interface](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/architecture/interface/gpio.hpp?ts=4#L167-L246).
```
ekiwi: plz tone down the teenager in you @salkinium (no `lol`)

PORTING.md
```diff
@@ -0,0 +1,327 @@
...
+
+Use the STM32 driver as a template, and modify it. For now you can ignore anything to do with interrupts, analog, alternate functions, `connect` methods, channels, triggers and ports.
+It is recommended that you implement the `configure`, `setOutput`, `setInput` methods for `InputType` and `OutputType`, so that you can put the GPIO into open drain mode for I2C.
+
+The elements from the device file is available as a name-mangled dictionary, with attributes available as their members.
```
ekiwi: This needs some clarification. Maybe:

> The children of the `<driver type="gpio" ...>` node are available as a name-mangled dictionary ...

PORTING.md
```diff
@@ -0,0 +1,327 @@
...
+```cpp
+extern "C"
+void
+xpcc_gpio_enable(void)
+{
+    // do your magic
+}
+```
+
+Remember, you can always check the generated code for this driver in `xpcc/build/sam4/empty/libxpcc/generated_platform/driver/gpio/sam4/`
```
ekiwi: I really like the fact that you are pointing people to the generated files as this is the way I debug my templates and also a serious advantage over preprocessor macros.

However, we have to make sure that people are including the header files that result from the templates, because `scons` will only generate the from the templates if the main target depends on them.

ekiwi:
Have a look at some of my thoughts.

All in all I think this guide is great and includes all that I can think of for porting `xpcc` to a new `cortex-m` based platform.

I am sure there will be some small things you (and now I) missed, but that will turn up, once someone uses this guide.

PORTING.md
salkinium: Well, I think both `defines()` and `appendInterrupt()` can be removed. We are using the CMSIS interrupt naming convention and that shouldn't change across platforms.

There is an issue with ordering, where I think 2. and 3. should be swapped. If the stack is placed in internal memory that needs to be enabled first, then the current solution won't work.

salkinium: In case the stack memory needs to be enabled first, the platform code must not use stack any memory, which when using C is not guaranteed to happen (the abstract C machine assumes availability of a stack and also initialized `.data` and `.bss` sections).
So the enablement code would then have to be written in assembly.

PORTING.md
salkinium: My intentions were for it to say:

``` cpp
// Use this hook to initialize external memories
xpcc_hook_hardware_init();
```

PORTING.md
salkinium: > inline the code from stm32.macros and lpc.macros?

No and yes. We should put the macro in the same file `cortex/platform.macros`. The right code is then chosen using filters inside the macro.
Keeps the startup-script focussed on Cortex-M, and makes it easier to add platforms.
I am also a bit annoyed that the linkerscripts are in the `cortex/stm32/` folder, instead of the `cortex/linkerscript` folder ;-P

PORTING.md
salkinium: That's an excellent idea, I'll rewrite this.

PORTING.md
salkinium: LOL.

PORTING.md
salkinium: "Porting the delay functions"? or "Enabling the delay functions"? I'd prefer porting, cos it's enabling is kinda obvious.

PORTING.md
salkinium: Shouldn't including the top level `<xpcc/architecture/platform.hpp>` be enough?

PORTING.md
ekiwi: > I am also a bit annoyed that the linkerscripts are in the cortex/stm32/ folder, instead of the cortex/linkerscript folder ;-P

what prevents you from moving the file?

PORTING.md
ekiwi: I like "Enabling the delay functions" if you can start the paragraph somehow like this:

> In order to enable the delay functions that come with xpcc on your platform, you need
> to provide some information about the frequency that the CPU core is running at.

PORTING.md
ekiwi: > Shouldn't including the top level <xpcc/architecture/platform.hpp> be enough?

Yes. However, I am afraid that people might just use an empty main ....

PORTING.md
```diff
@@ -0,0 +1,351 @@
...
+    // Don't forget to enable the GPIO clock!
+    // GPIO_PORTA_OUTPUT = (1 << 0);
+
+    while(1)
+    {
+        // Toggle the pin manually
+        // GPIO_PORTA_TOGGLE = (1 << 0);
+
+        // create a fake busy wait
+        for (uint32_t ii = 0; ii < 1000000; ii++) ;
```
ekiwi: this can be dangerous because for two reasons:
1. compiler optimizations: how about we include some `volatile asm "nop"` or similar?
2. if people have no intuition for how fast their system is executing the delay could be unnoticeabley short or it could be so long that people don't see it: maybe we should provide some guidance on how to chose the loop bound

PORTING.md
salkinium: I'm preparing a PR right now with those changes.

PORTING.md
```diff
@@ -0,0 +1,351 @@
...
+}
+}
+```
+
+If you [look at the STM32 `driver.xml` file](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/architecture/platform/driver/clock/stm32/driver.xml#L19-L22), you'll see that the internal high- and low-speed oscillator values differ even within the same platform. These values are therefore initialized in the clock driver and not the core driver.
+
+### Porting the GPIO driver
+
+> In xpcc, GPIOs are assumed to be usable without explicitly initializing the subsystem.
+> For other peripheral IP, the clock is enabled in the `initialize` method that needs to be explicitly called by the user.
```
ekiwi: That wasn't intended to be included as a quote. I only used the quote style in order to visually distinguish the parts that I would included in the document from my general comments.

salkinium: Ah, I thought it was port of our documentation and therefore quoted. Will inline then.

PORTING.md
salkinium: > Yes. However, I am afraid that people might just use an empty main ....

We can only hope they have some valid healthcare when they shoot themselves in the foot.

salkinium:
Updated for latest review.

PORTING.md
```diff
@@ -0,0 +1,354 @@
...
+
+## Porting a new device from AVR or STM32
+
+It may be as easy as adding a new device file, but if it's a brand new device, the device headers also need to be updated. When porting a new series (for example the STM32L4), a lot of code can be reused, but there may be very important differences between the two that are not immediately obvious to someone who isn't that familiar with STM32 devices.
+
+We recommend writing us an email, or opening up an issue on GitHub.
+
+## Porting a new device from a new vendor
+
+Note: We assume this new device is Cortex-M based. For other CPU architectures, send us an email.
```
ekiwi: this needs a link to the mailing list

PORTING.md
```diff
@@ -0,0 +1,354 @@
+# Porting to xpcc
+
+## What xpcc supports
```
ekiwi: > What xpcc already supports
> ?

PORTING.md
```diff
@@ -0,0 +1,354 @@
...
+
+- AT90, ATtiny, ATmega devices
+- STM32 F0, F1, F3, F4, F7 devices.
+
+If your device falls into any of those, you can just use the specific device
+identifier in your `project.cfg` and it'll very likely just work out-of-the-box.
+
+If you want to add a development board for an existing device, have a look at [`xpcc/architecture/platform/board`](https://github.com/roboterclubaachen/xpcc/tree/develop/src/xpcc/architecture/platform/board).
+
+## Porting a new device from AVR or STM32
```
ekiwi: > Adding support for new STM32 or AVR devices

PORTING.md
```diff
@@ -0,0 +1,354 @@
...
+
+It may be as easy as adding a new device file, but if it's a brand new device, the device headers also need to be updated. When porting a new series (for example the STM32L4), a lot of code can be reused, but there may be very important differences between the two that are not immediately obvious to someone who isn't that familiar with STM32 devices.
+
+We recommend writing us an email, or opening up an issue on GitHub.
```
ekiwi: here I would also like to include a link to the mailing list as well as to the github issue page

PORTING.md
```diff
@@ -0,0 +1,354 @@
...
+
+Porting to a new vendor requires you to:
+
+1. Add a bunch of header files for the new device.
+2. Add a device files for the new device.
+3. Make the build system aware of the new device classes.
+4. Adapt the Cortex-M startup- and linkerscripts.
+5. Compile a minimal example.
+6. Write a basic GPIO and delay driver.
+7. Compile an LED blinking example.
```
ekiwi, line 33: This should be a table of content with hyperlinks, because every time I see this I think that this is the only information that I will get only to discover that there is more information provided below.
However, apparently there is no support for automatic tables of content in github markdown.
The only solution seems to be to manually create the table of content or to switch to [AsciiDoctor](https://medium.com/@chacon/living-the-future-of-technical-writing-2f368bd0a272)

PORTING.md
```diff
@@ -0,0 +1,354 @@
...
+
+### Adding device header files
+
+Copy the [CMSIS header files](https://www.arm.com/products/processors/cortex-m/cortex-microcontroller-software-interface-standard.php)  for the device family into the `ext/` folder. You can usually download them as part of a driver pack from the vendor's website.
+Have a look at [the STM32 device headers](https://github.com/roboterclubaachen/xpcc/tree/develop/ext/st) as an example for what you are looking for.
+
+Create a `README.md` that contains the source and copyright (!) of the header files.
```
ekiwi: yes! And please include the the license and source of the vendor files in your pull request.

PORTING.md
```diff
@@ -0,0 +1,354 @@
...
+Don't forget to add the include path of the device files to the [`ext/SConstruct`](https://github.com/roboterclubaachen/xpcc/blob/develop/ext/SConscript) file!
+
+
+### Adding device files
+
+The [xpcc device files](https://github.com/roboterclubaachen/xpcc/tree/develop/src/xpcc/architecture/platform/devices) contain the information that is used to generate the HAL. For AVR and STM32 devices these files are generated from another (vendor specific) data source.
+However, for a new vendor such machine readable data is likely not available, and we recommend building a minimal proof-of-concept port for one device first.
```
ekiwi: I feel like the actual problem is not availability of machine readable data, but rather that it would be overkill at this stage to actually write a parser to convert this data to xpcc device files.

PORTING.md
salkinium: I'll just make it clearer in the text that this is an overview. I'm too old to battle Markdown. #LifeCrisis

PORTING.md
```diff
@@ -0,0 +1,354 @@
...
+
+You need to manually create a new device file for your device. We recommend copying and modifying [this device file for the STM32F303](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/architecture/platform/devices/stm32/stm32f303-c_k_r-6_8.xml) for your device.
+
+Vendors usually have a naming convention of their devices, make sure the device attributes and file name reflect this.
+For example, the `SAM4LC4B` identifier can be split like this:
+```xml
+  <device platform="sam" family="4" name="lc" pin_id="b" size_id="4">
+    <!-- ... -->
+  </device>
+```
```
ekiwi, line 61: > The semantics of the vendors device naming scheme can sometimes be gathered directly from the datasheet
> or by comparing different devices (i.e. their names and capabilities).

PORTING.md
```diff
@@ -0,0 +1,354 @@
...
+
+Modify the F303 device file to reflect the new devices data.
```
ekiwi: argh. This sound almost like "Modify the F303 device file _in place_ ..." which is not what you are intended to do here.

PORTING.md
```diff
@@ -0,0 +1,354 @@
...
+Since you only have one specific device to port, you can throw out the duplicate XML tags which are normally filtered by `device-` attributes.
+
+Adding the FLASH and RAM size allows the build system to display accurate static resource usage:
```
ekiwi: > ... and to warn you if your program does not fit into memory.

PORTING.md
```diff
@@ -0,0 +1,354 @@
...
+```xml
+<flash>65536</flash>
+<ram>16384</ram>
+```
+
+Choose the correct ARM Cortex-M core (`m0`, ~~`m0p`~~, `m3`, `m4`, `m4f`, `m7`, `m7f`, `m7fd`):
```
ekiwi: Do we support the `cortex-m0p` device? (why not `cortex-m0+`?)
Either we add support for this (it does not need to be treated differently to `m0`, only add it to the check so that there is no error, when using it), or we don't include it in this list.

PORTING.md
```diff
@@ -0,0 +1,354 @@
...
+```xml
+<core>cortex-m4f</core>
+```
+
+And correct CMSIS device header and device macro:
+```xml
+<header>sam4.h</header>
```
ekiwi, line 79: please add some information on how this file is discovered through the include path, i.e. how this is connected to modifying the `SConscript` in the `ext` folder

PORTING.md
```diff
@@ -0,0 +1,354 @@
...
+<define>SAM4LC4B</define>
```
ekiwi, line 84: Do we really want to include this for new devices?
Going forward, it would be nice if we could just rely on our `jinja2` templates instead of involving the C preprocessor.

PORTING.md
```diff
@@ -0,0 +1,354 @@
...
+```
+
+You need to implement at least three drivers: `core`, `clock` and `gpio`.
```
ekiwi: We aren't really "implementing" the drivers here, more like "instantiating" or "including"

PORTING.md
salkinium: > Do we support the cortex-m0p device?

No, because we've never had to.

> (why not cortex-m0+?)

I'm afraid of the `+` causing problems somewhere.

> Either we add support for this

For which device can we test this?

> or we don't include it in this list.

Seems better for now.

PORTING.md
```diff
@@ -0,0 +1,354 @@
...
+
+Make sure you give the `core` driver the right information about the memory map. This information is used to generate the linkerscript.
+Don't worry about different memories right now, just add the main FLASH and one SRAM region here (called `sram1`).
+```xml
+<driver type="core" name="cortex">
+  <memory access="rx" start="0x8000000" name="flash" size="64"/>
+  <memory access="rwx" start="0x20000000" name="sram1" size="12"/>
```
ekiwi, line 94: What is the SI unit of `size`?

PORTING.md
```diff
@@ -0,0 +1,354 @@
...
+The `name` will be appended with `_Handler`, so declaring `extern "C" void WWDG_Handler(void) {...}` anywhere will link that function into the vector table (note the C linkage!).
+```xml
+  <vector position="0" name="WWDG"/>
+  <vector position="1" name="PVD"/>
+  <!-- ... -->
+  <vector position="81" name="FPU"/>
+</driver>
+```
+
+The `clock` driver just needs to be present, we'll hardcode the specifics in code:
```
ekiwi: again I would prefer if the driver is `included` or `instantiated`

PORTING.md
```diff
@@ -0,0 +1,354 @@
...
+# path to the xpcc root directory
+xpccpath = '../../..'
+# execute the common SConstruct file
+execfile(xpccpath + '/scons/SConstruct')
+```
+
+- an empty `main.cpp` file:
+```cpp
+// your copyright with bsd license
+#include <xpcc/architecture/platform.hpp>
```
ekiwi, line 181: ok, since you mention this include here we should be good

PORTING.md
```diff
@@ -0,0 +1,354 @@
...
+```python
+# path to the xpcc root directory
+xpccpath = '../../..'
+# execute the common SConstruct file
+execfile(xpccpath + '/scons/SConstruct')
+```
+
+- an empty `main.cpp` file:
+```cpp
+// your copyright with bsd license
```
ekiwi: why not just include the short BSD license header that we use, including the Copyright template with `{{ your_name }}`?

I always find it tedious to get hold of this.

PORTING.md
salkinium: It's not our call, at least not for ST, since the CMSIS headers choose their headers and implementations based on this. There is no clear pattern for predicting the macro though, [hence this table exists](https://github.com/roboterclubaachen/xpcc/blob/develop/tools/device_file_generator/stm.py#L9-L104).

PORTING.md
```diff
@@ -0,0 +1,354 @@
...
+    while(1)
+    {
+        // Toggle the pin manually
+        // GPIO_PORTA_TOGGLE = (1 << 0);
+
+        // Create a very crude busy wait. The number don't have to
+        // be super accurate, just in the ball-park of a 1s delay.
+        // Use the initial CPU frequency in MHz here.
+        constexpr uint32_t fcpu_MHz = 8;
+        for (volatile uint32_t ii = 0; ii < fcpu_MHz * 250000; ii++) ;
```
ekiwi, line 201: are you 100% positive, that this loop is not optimized away?

PORTING.md
ekiwi: OK. Maybe just add the information that this is required by `CMSIS` in your guide so that people know what macro they need to define if they are porting to a different platform.

PORTING.md
```diff
@@ -0,0 +1,354 @@
...
+(.data + .reset + .rodata + .text)
+
+Data:       1000 bytes (2.0% used) = 1 bytes static (1.0%) + 1000 bytes stack (1.0%)
+(.bss + .data + .noinit + .stack)
+
+Heap:      9999 bytes (97.0% available)
+(.heap1)
+```
+
+You can explore the generated code for the core driver in `xpcc/build/sam4/minimal/libxpcc/generated_platform/driver/core/cortex/`.
```
ekiwi, line 233: +1

PORTING.md
```diff
@@ -0,0 +1,354 @@
...
+```
+For custom setups, [read this](http://xpcc.io/reference/build-system/#project-configuration).
+
+You should now be able to program the device using `scons program`.
+Check if the LED is blinking at a constant interval.
+
+If the LED is not blinking and you have OpenOCD connected, you can [launch a GDB debug session](https://github.com/roboterclubaachen/xpcc/tree/develop/examples/stm32f3_discovery/gdb).
+
+If something goes wrong in the building step, check the previous steps, maybe you missed something, or we forgot to add something to this guide.
+If you're totally stuck write us an email (try to work the problem first though ;-).
```
ekiwi: Include link to mailing list.
Also I think the last part could be understood to be condescending and should be clear to anyone using this guide. I'd just drop it.

PORTING.md
```diff
@@ -0,0 +1,354 @@
...
+
+    return 0;
+}
+```
+
+After compilation and programming, the the LED should now be blinking at a 1 second interval.
+
+### What's next?
+
+Please create a PR with this work, so we can review it and discuss the next steps.
```
ekiwi, line 377: > Please include all the sources that you used for your port as well as the associated licenses. All code that does not 
> reside in `ext` must be licensed under BSD license.

salkinium:
Updated for review.

PORTING.md
```diff
@@ -0,0 +1,384 @@
+# Porting to xpcc
+
+## What xpcc already supports
+
+xpcc contains extensive support for AVR and STM32 devices, specifically:
+
+- AT90, ATtiny, ATmega devices
+- STM32 F0, F1, F3, F4, F7 devices.
+
+If your device falls into any of those, you can just use the specific device
```
ekiwi: > [...] any of those categories,  you can [...]

PORTING.md
```diff
@@ -0,0 +1,384 @@
...
+Have a look at [the STM32 device headers](https://github.com/roboterclubaachen/xpcc/tree/develop/ext/st) as an example for what you are looking for.
+
+Create a `README.md` that contains information on the source of the header files, with a link to the vendor website, the download date/version and most importantly the license text.
+Don't forget to add the include path of the device files to the [`ext/SConstruct`](https://github.com/roboterclubaachen/xpcc/blob/develop/ext/SConscript) file!
+
+
+### Adding device files
+
+The [xpcc device files](https://github.com/roboterclubaachen/xpcc/tree/develop/src/xpcc/architecture/platform/devices) contain the information that is used to generate the HAL. For AVR and STM32 devices these files are generated from another (vendor specific) data source.
+However, for a new vendor such machine readable data is likely not available and if it were, it's likely too much work to convert it automatically to our device file format.
```
ekiwi: > However, for a new vendor such machine readable data might not be available and if it is, it will require some work to create the necessary parser to automatically generate device files.
> We recommend building a minimal proof-of-concept port for one device first.
> Once you have one device working, you can get started working on an automatic parser if you want to support a wide range of devices.

PORTING.md
```diff
@@ -0,0 +1,384 @@
...
+We recommend building a minimal proof-of-concept port for one device first.
+
+You need to manually create a new device file for your device. We recommend copying [this device file for the STM32F303](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/architecture/platform/devices/stm32/stm32f303-c_k_r-6_8.xml) to a new folder named after the platform (like `devices/sam`).
+
+Vendors usually have a naming convention of their devices, make sure the device attributes and file name reflect this. The semantics of the vendors device naming scheme can sometimes be gathered directly from the datasheet or by comparing different devices (i.e. their names and capabilities).
```
ekiwi: > vendor's

PORTING.md
```diff
@@ -0,0 +1,384 @@
...
+For example, the `SAM4LC4B` identifier can be split like this:
+```xml
+  <device platform="sam" family="4" name="lc" pin_id="b" size_id="4">
+    <!-- ... -->
+  </device>
+```
+
+Modify the copied F303 device file to reflect the new devices data.
```
ekiwi: > the copied device file

PORTING.md
```diff
@@ -0,0 +1,384 @@
...
+```
+
+- a minimal `main.cpp` file:
+```cpp
+/* Copyright (c) {{ year }}, {{ your name }}
+ * All Rights Reserved.
+ *
+ * The file is part of the xpcc library and is released under the 3-clause BSD
+ * license. See the file `LICENSE` for the full license governing this code.
+ */
```
ekiwi, line 180: +1

ekiwi:
once my comments are addressed I would be happy to merge this!

salkinium:
Addressed review, however this PR will change slightly once #180 is merged.

ekiwi:
Nice! Please merge. We can adjust this later, when #180 is merged.

salkinium:
Thanks for the thorough review, @ekiwi!

daniel-k:
@salkinium Can you add link to this porting guide to the xpcc.io page? It's a bit too hidden, only found it because I remembered there had been this PR a while ago.

salkinium:
Done, both on website and on repo readme.

daniel-k:
It's always a pleasure :smiley:

PR #188, opened 2016-09-20
Deduplicate and clean up linker scripts

salkinium:
Deduplication by using Jinja2 macros to deduplicate the linker script formatting followed by clean up of unused or pointless linker sections.

I believe most of the original linker script code comes from @dergraaf. Please review the removal of the sections, in case I missed something.

Note: This requires review and testing in hardware.

cc @ekiwi @strongly-typed.


salkinium:
This has now been tested in hardware on all supported platforms.

PR #189, opened 2016-09-22
[DFG] Cleanup python code and add naming schema to device files

dergraaf:
Uses local imports in the Python code and therefore removes the need to update the python path in every file.

Adds a 'naming-schema' and 'invalid-device' tag to the device files to allow to enumerate all valid device names from the device files alone. Removes the size_id attribute for AVRs because it is redundant with the name and not used by the AVR DFG merger.


salkinium:
:+1: for naming schema, so cool! Also thanks for cleaning up my Python mess!

I'm split about the `invalid-device` blacklist. I see how it enables the device file to be self contained and how it's a more compact representation than a `valid-device` whitelist in most cases. I'll have to think about that some more, I can't think of a better alternative right now.

salkinium (approved):

PR #196, opened 2016-10-14
[scons] Add compiler optimization command line argument.

rleh:
Sometime useful for debugging.

### DIFF
--- a/scons/site_tools/arm.py
+++ b/scons/site_tools/arm.py
@@ -152,6 +152,7 @@ def generate(env, **kw):
 		"-DBASENAME=${SOURCE.file}",
 		"-fdiagnostics-color=auto",
 	]
+	env['CCFLAGS'] += env['XPCC_ADDITIONAL_CCFLAGS']
 
 	# Platforms with single precision float support
 	if env['ARCHITECTURE'] in ['cortex-m4f', 'cortex-m7f']:
--- a/scons/site_tools/arm.py
+++ b/scons/site_tools/arm.py
@@ -124,7 +124,7 @@ def generate(env, **kw):
 		"$FPU",
 		"$THUMB",			# use THUMB='-mthumb' to compile as thumb code (default for AT91SAM)
 		"-mthumb-interwork",
-		"-Os",
+		"$OPTIMIZATION",
 		"-gdwarf-2",
 		"-funsigned-char",
 		"-funsigned-bitfields",
--- a/scons/site_tools/avr.py
+++ b/scons/site_tools/avr.py
@@ -140,6 +140,7 @@ def generate(env, **kw):
 		"-Wa,-adhlns=${TARGET.base}.lst",
 		"-DBASENAME=${SOURCE.file}",
 	]
+	env['CCFLAGS'] += env['XPCC_ADDITIONAL_CCFLAGS']
 	
 	# C++ flags
 	env['CXXFLAGS'] = [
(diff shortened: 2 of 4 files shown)


ekiwi:
Good idea. I normally just overwrite the flag in my `SConstruct` file but that is a suboptimal solution, as I first have to remove the default set by xpcc.

I am not sure if the solution you propose is ideal though. How about adding the option to set optimization level and additional flags (like `-Werror`) in the `project.cfg`.
@salkinium ?

rleh:
[I found your gdb tutorial for the STM32F3 Discovery](https://github.com/roboterclubaachen/xpcc/blob/develop/examples/stm32f3_discovery/gdb/Readme.md)

I don't like the idea of changing a file to set the compiler optimization very much.

E.g. if you use QtCreator you could disable optimization for debug builds while using optimization in the normal builds.

salkinium:
> E.g. if you use QtCreator you could disable optimization for debug builds while using optimization in the normal builds

Yes, this is exactly the reason why you would to specify this on the command line.

However, it may be useful to be able to specify an optimization level in the `project.cfg`, which gets overwritten by the command line option. This way you can share optimization levels with other people. But I don't feel it is required, this change is already very useful.

rleh:
Ok, good idea.
I'll add a `project.cfg` option.

salkinium:
I like this idea, but it's implemented a bit weirdly. `avr.py` and `arm.py` should not be duplicating the same argument parsing, [`xpcc.py` is the place to do that](https://github.com/roboterclubaachen/xpcc/blob/develop/scons/site_tools/xpcc.py#L313). You also need to add an optimization level to `hosted.py`.

Note that `get()` can be provided with a default:

``` py
    # compiler optimization
    optimization_levels = ['0', '1', '2', '3', 's']
    optimization = ARGUMENTS.get('optimization', 's')
    if optimization is in optimizaton_levels:
        env['OPTIMIZATION'] = "-O{}".format(optimization)
    else:
        env.Error("Unknown optimization level: '{}'".format(optimization))
        Exit(1)
```

salkinium:
You can also do the `project.cfg` option parsing and overwriting much easier in `xpcc.py`.

ekiwi:
How about being able to specify different build configurations in the project.cfg and then choosing from them through a scons argument?

The problem that I have with the optimization argument is that it only covers one use case.

rleh:
> The problem that I have with the optimization argument is that it only covers one use case.

In my opinion it is sufficient for the most use cases to disable the compiler optimization while debugging.

The optimization argument is available for every xpcc project without modifications to the `project.cfg`.

Independent of this we might add another option to the `project.cfg` to specify additional CCFLAGS like this:

``` ini
[build]
# ...
ccflags = '-Werror -Wall -Wextra'
```

ekiwi:
Just looked at your code and it seems good to me.
Did you intentionally ignore hosted?

rleh:
>  Did you intentionally ignore hosted?

[No.](https://github.com/roboterclubaachen/xpcc/pull/196/commits/c84a556bd24db4294e9432bc8ceed98ec5e80921)

salkinium:
I've added getting the optimization level via cascading defaults which makes the code easier to read. You can rebase this and make my commits disappear (or I can do it) if you like these changes.

salkinium:
Note to self: Update website with new `project.cfg` and command line settings!

salkinium (approved):

salkinium:
Thanks!

PR #197, opened 2016-10-15
Miscellaneous fixes

salkinium:
Didn't want to open a PR for each of these small fixes.

cc @ekiwi.

### DIFF
--- a/src/xpcc/architecture/driver/accessor/flash_reader_avr_impl.hpp
+++ b/src/xpcc/architecture/driver/accessor/flash_reader_avr_impl.hpp
@@ -41,7 +41,7 @@
 #define	FLASH_STORAGE_STRING(s)			extern const char s[] PROGMEM; const char s[]
 #define	EXTERN_FLASH_STORAGE_STRING(s)	extern const char s[] PROGMEM
 
-#define	INLINE_FLASH_STORAGE_STRING(s)	((const PROGMEM char *)(s))
+#define	INLINE_FLASH_STORAGE_STRING(s)	PSTR(s)
 
 namespace xpcc
 {
--- a/scons/site_tools/avrdude.py
+++ b/scons/site_tools/avrdude.py
@@ -31,7 +31,7 @@ from SCons.Script import *
 
 # -----------------------------------------------------------------------------
 def avrdude_flash(env, source, eeprom_source='', alias='avrdude_program'):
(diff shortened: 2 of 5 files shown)


src/xpcc/architecture/driver/accessor/flash_reader_avr_impl.hpp
```diff
@@ -41,7 +41,7 @@
 #define	FLASH_STORAGE_STRING(s)			extern const char s[] PROGMEM; const char s[]
 #define	EXTERN_FLASH_STORAGE_STRING(s)	extern const char s[] PROGMEM
 
-#define	INLINE_FLASH_STORAGE_STRING(s)	((const PROGMEM char *)(s))
+#define	INLINE_FLASH_STORAGE_STRING(s)	PSTR(s)
```
ekiwi, line 44: is the `PROGMEM` attribute not needed here?

salkinium: No, for AVRs `PSTR` is provided by `avr-libc`.

ekiwi: ok, but why did you have to define it for other targets then?
I am assuming that `flash_reader_avr_impl.hpp` is only included for AVR targets, right?

salkinium: It's a replacement for `INLINE_FLASH_STORAGE_STRING`, so you can compile code for AVRs also on ARM Cortex-M.

ekiwi: but wasn't that the purpose of the `INLINE_FLASH_STORAGE_STRING` define, to provide a platform independent option?

salkinium: Yes, but:

``` cpp
XPCC_LOG_DEBUG << xpcc::accessor::asFlash(INLINE_FLASH_STORAGE_STRING("Temperature is ")) << t << xpcc::accessor::asFlash(INLINE_FLASH_STORAGE_STRING(" inside kitchen."));
```

OMGWTFBBQ was I smoking when I thought that `INLINE_FLASH_STORAGE_STRING` is an acceptable length for a macro that is to be used inline _a lot_?

ekiwi: hm, but `PSTR` isn't descriptive at all...

salkinium: No, but people coming from AVR will know it, and it won't break ARM Cortex-M this way.

PR #199, opened 2016-11-14
Import OPENOCD_SCRIPTS env variable to OpenOCD

7Kronos:
As described in the OpenOCD documentation :
http://openocd.org/doc/html/Running.html

It is possible to specify the OpenOCD scripts directory path by setting
an environment variable. This is particulary handy for Windows users
(specialy if the scons command is executed within Visual Studio.

### DIFF
--- a/scons/site_tools/openocd.py
+++ b/scons/site_tools/openocd.py
@@ -44,6 +44,10 @@ def openocd_run(env, source, alias='openocd_run'):
 	if platform.system() == "Windows":
 		env['OPENOCD_COMMANDS'] = env['OPENOCD_COMMANDS'].replace("$SOURCE", str(source[0]).replace("\\","/"))
 
+	# Provide additional search paths via the OPENOCD_SCRIPTS environment variable
+	# See http://openocd.org/doc/html/Running.html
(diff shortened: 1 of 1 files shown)


salkinium:
I did not know this about OpenOCD, this is a very good idea!
However, the current code will raise a `KeyError` if the variable is not set in the environment.

This code works for me on OS X with and without the variable set:

``` python
env['ENV']['OPENOCD_SCRIPTS'] = os.environ.get('OPENOCD_SCRIPTS')
```

`os.environ.get` returns `None` if the variable is not set and setting a SCons environment variable to `None` should not modify it, so this should always work.

7Kronos:
Hi, sorry I'm a total newbee to Python :)
Fixed!

salkinium:
> Hi, sorry I'm a total newbee to Python :)

No worries! Thanks, I'll squash and merge this when I'm back from work.

salkinium:
I made a minor change during my squashing, so that this functionality is available for any OS, not just Windows. Thanks for this!

I pushed to your develop branch, which is suboptimal for you, sorry. You may have to `git fetch; git reset --hard origin/develop` on your local develop branch (make sure you don't have local changes).
I recommend you start a feature branch for every change, so that you can always update your develop branch with the upstream changes.

PR #210, opened 2017-01-23
Fix ZeroMQ communication backend

chris-durand:
This fixes several problems in the ZeroMQ backend. The header-only implementation is also moved to cpp files. Thus ZeroMQ and zmqpp will be required for building xpcc on hosted. Is this a problem?

### DIFF
--- a/src/xpcc/communication/xpcc/backend/zeromq/reader.hpp
+++ b/src/xpcc/communication/xpcc/backend/zeromq/reader.hpp
@@ -1,0 +1,177 @@
+// coding: utf-8
+/* Copyright (c) 2016, Roboterclub Aachen e. V.
+ * All Rights Reserved.
+ *
+ * The file is part of the xpcc library and is released under the 3-clause BSD
+ * license. See the file `LICENSE` for the full license governing this code.
+ */
+// ----------------------------------------------------------------------------
+
+#ifndef	XPCC__ZEROMQ_READER_HPP
+#define	XPCC__ZEROMQ_READER_HPP
+
+#include <thread>
+#include <mutex>
+#include <atomic>
+#include <deque>
+
+#include <zmqpp/zmqpp.hpp>
+
+#include <xpcc/debug/logger.hpp>
+#undef XPCC_LOG_LEVEL
+#define	XPCC_LOG_LEVEL xpcc::log::ERROR
+
+namespace xpcc
+{
+
+class ZeroMQReader
+{
+public:
+	static constexpr int PollTimeoutMs = 100;
+
+	struct Packet
+	{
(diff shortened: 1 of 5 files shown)


salkinium (approved):

salkinium:
> Thus ZeroMQ and zmqpp will be required for building xpcc on hosted. Is this a problem?

That may be ok, but better ask @strongly-typed.

dergraaf:
@chris-durand Can you update the files with correct copyright information (not "Roboterclub Aachen" but real persons)?

chris-durand:
> That may be ok, but better ask @strongly-typed 

He suggested me to add a unit test which would require ZeroMQ as a build dependency anyway.

chris-durand:
> @chris-durand Can you update the files with correct copyright information (not "Roboterclub Aachen" but real persons)?

I added comments with author information to the header files.

src/xpcc/communication/xpcc/backend/zeromq/reader.hpp
```diff
@@ -0,0 +1,177 @@
+// coding: utf-8
+/* Copyright (c) 2016, Roboterclub Aachen e. V.
```
ekiwi: @chris-durand this needs to be the actual person who created this code, German copyright law afaik does not allow you to assign copyright to a company

chris-durand: @ekiwi This line is in every source file of xpcc, I have just copied that.

ekiwi: yes, we recently discovered, that this is non ideal

this is also what @dergraaf was referring to in his comment about updating the copyright information.

salkinium: This should really be:
```diff
-/* Copyright (c) 2016, Roboterclub Aachen e. V.
+/* Copyright (c) 2016, Sascha Schade
+ * Copyright (c) 2017, Christopher Durand
```
We are going to [forward port these changes to modm](https://github.com/modm-io/modm/blob/develop/src/modm/communication/xpcc/backend/zeromq/connector.hpp#L2), which use the correct copyright information. So to make it a little easier for us, we are requesting this change.

salkinium:
We are currently working on the next generation of this framework (over at [modm-io](https://github.com/modm-io), it's very much WIP), which will be licensed under MPLv2 (see below for why). Since this is a non-trivial change to xpcc, would you be ok with licensing your contributions as MPLv2 as well?

> 
>

chris-durand:
@salkinium I have added the requested copyright information and am ok with relicensing my contributions as MPLv2.

PR #216, opened 2017-02-19
[resumable] Add nesting depth assertion.

salkinium:
Add an assertion to catch nesting depth exhaustion for nested resumable functions.

### DIFF
--- a/src/xpcc/architecture/interface/assert.hpp
+++ b/src/xpcc/architecture/interface/assert.hpp
@@ -155,7 +155,8 @@ xpcc_abandon(const char * module,
 #endif
 
 #define xpcc_assert(condition, module, location, failure) \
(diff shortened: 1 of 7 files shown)


salkinium:
wat? The unittests fail with
```
driver/core/hosted/assert.cpp:37: undefined reference to `__start_xpcc_assertion'
driver/core/hosted/assert.cpp:38: undefined reference to `__stop_xpcc_assertion'
```
but [examples/linux/assert compiles fine](https://travis-ci.org/roboterclubaachen/xpcc/jobs/203225595#L1539)? Le sigh.

salkinium:
This makes sense, since for the unittest no handler has been declared and so the section `xpcc_assertion` doesn't exist. I guess I'm not using the dynamic linker correctly.

PR #218, opened 2017-03-04
[unittest] BoundedDeque: Add single element test

tomchy:
I want to use single element queue, but it doesn't work. See unit test in this PR.

@daniel-k told me to do so :slightly_smiling_face:

### DIFF
--- a/src/xpcc/container/deque_impl.hpp
+++ b/src/xpcc/container/deque_impl.hpp
@@ -36,7 +36,7 @@
 
 template<typename T, std::size_t N>
 xpcc::BoundedDeque<T, N>::BoundedDeque() : 
-	head(0), tail(1), size(0)
+	head(0), tail(N == 1 ? 0 : 1), size(0)
 {
 	static_assert(N > 0, "size = 0 is not allowed");
 }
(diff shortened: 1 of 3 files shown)


tomchy:
It seems that changing static assignment of tail fixes this issue (still should be checked):
```c
this->tail = (N == 1 ? 0 : 1);
```

This assignment can be found inside constructor and clear() method.

salkinium:
Is this PR missing the commit with the fix?

tomchy:
Yes. I've just pushed a simple fix. Please check :slightly_smiling_face:

src/xpcc/container/deque_impl.hpp
```diff
@@ -36,8 +36,9 @@
 
 template<typename T, std::size_t N>
 xpcc::BoundedDeque<T, N>::BoundedDeque() : 
-	head(0), tail(1), size(0)
+	head(0), size(0)
```
salkinium: I'd like this still in the constructor list, that will likely result in one less constructor call for static instantiations.
```diff
-	head(0), tail(1), size(0)
+	head(0), tail(N == 1 ? 0 : 1), size(0)
```

salkinium (changes requested):
Apart from the minor change, this is 👍.

tomchy:
Changes applied :slightly_smiling_face:

salkinium (approved):

daniel-k:
👍

salkinium:
I've squashed and rebased your commit for ~yak~ git history shaving purposes.

salkinium:
Thanks a lot for the fix!

PR #221, opened 2017-03-10
Add NUCLEO-F031K6 board and example

genbattle:
Hi, I added support for the Nucleo F031k6 board to allow me to develop for it using XPCC. So far I've developed the platform definition based on the existing definition for the f103rb device (including support for other devices in the f031xx family). I've got the platform working with the basic blink example.

Still to be done:

- [x] OpenOCD support; there's no file for the nucleo f031k6 included with OpenOCD currently, so I have developed my own configuration file for this platform that seems to work, but I'm not sure where it fits in within XPCC.
- [x] Tests and examples for this platform.
- [x] Verify serial debugging works properly.

I would appreciate any guidance you can give me on whether this work will be valuable to the project, and how it might be improved (especially with respect to adding OpenOCD support for the platform) before it is merged.

PR #224, opened 2017-03-15
Add lost coding conventions

### LINKED ISSUE #208: Coding conventions are missing
The coding conventions referenced from the contribution guidelines link to the old `xpcc-doc` repo, which we replaced with the `xpcc.io` repo (which doesn't contain the conventions either, lol).

Here is the last known version of the conventions:
https://gist.github.com/salkinium/3dba5ef2570a22ba76c748ee0c8243ad

salkinium:
Adds back the coding conventions (fixes #208).

~~Adds documentation to deprecate the DFG in favor of the one in [modm-devices](https://github.com/modm-io/modm-devices).~~
~~Adds all available device files to avoid future manual device file porting (relates to #221).~~
Moved to #226.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,9 +7,7 @@ We recommend prefixing your branch with
 - `feature/` for new functionality, or
 - `fix/` for a bug fix.
 
-Please remember to have a look at [our](https://github.com/roboterclubaachen/xpcc-doc/blob/master/source/developer/coding_conventions.rst)
-[coding](https://github.com/roboterclubaachen/xpcc-doc/blob/master/source/developer/coding_conventions.cpp)
-[conventions](https://github.com/roboterclubaachen/xpcc-doc/blob/master/source/developer/coding_conventions.hpp).
+Please remember to have a look at [our coding conventions](https://github.com/roboterclubaachen/xpcc/blob/develop/doc/coding_conventions.md).
 
 Once you [open the pull request](https://github.com/roboterclubaachen/xpcc/pulls), we will review it and might ask you for changes.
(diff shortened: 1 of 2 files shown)


strongly-typed:
Could you please also add STM32Lxx device files? You [added one here ](https://github.com/salkinium/xpcc/commit/47949a8fd0e734356a3c220674119ba1a50f5216) for my greatest pleasure. I just [resurrected this branch](https://github.com/strongly-typed/xpcc/commits/feature/stm32l4_support_2017) (and stripped the hwut) because will be discussing using STM32L476 with xpcc tomorrow with my team (after being disappointed multiple times by ST Cube HELL).

salkinium:
> Could you please also add STM32Lxx device files?

Not in this PR. There are a number of things that need to be added and modified, so I'd like to see your branch as a separate PR.

Keep in mind, that the DFG in this repo generates inaccurate device files, since the package id is missing. This is ok, as long as you double check the device file for your specific device. The modm DFG has been rewritten from scratch to prevent these issues from the start. Unfortunately the device file format isn't backwards compatible.

strongly-typed:
> so I'd like to see your branch as a separate PR.

Is it already worth a PR? Only some GPIOs and UART are tested, but not much more.

salkinium:
Hm, then let's make a PR with just the lower part of that branch.

salkinium:
Note that there are a few devices missing from this, in particular: F413, F423, F722, F723, F732, F733. This is because the device header files and some meta information have not been updated yet and the DFG depends on that.

salkinium:
I'm changing this PR to only deal with the coding convention. I'll add more substantial changes to how we deal with device files and device headers in #226.

salkinium:
Any objections for merging this?

PR #227, opened 2017-03-18
Driver: Adapt SSD1306 to 128 x 32 pixels display.

strongly-typed:
128x64 pixel resolution is hard-coded in driver.

There are also displays with 128x32 resolution available. This code is successfully tested with such a display.

What is the best approach to support these two displays? Enum Class `Resolution` as template parameter?
I will update the PR asap.


salkinium:
Make it an integer, but limit it with a static assert to either 64 or 32 for now. I think that's the best solution.
```c++
template < class I2cMaster, uint16_t  Height = 64 >
class Ssd1306 : public ssd1306, public BufferedGraphicDisplay<128, Height>
```

strongly-typed:
Humble question:
Does [buffer](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/driver/display/ssd1306.cpp#L58) refer to [buffer](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/ui/display/buffered_graphic_display.hpp#L107) or to [buffer](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/architecture/interface/i2c_transaction.hpp#L66)?
<strike>Interestingly</strike>, when renaming `buffer` to `pixbuf` in `BufferedGraphicDisplay` the existing code <strike>[breaks](https://github.com/strongly-typed/xpcc/tree/fix/buffer)</strike>.

strongly-typed:
The underlying problem seems to be that when `ssd1306.cpp` is compiled the display buffer size must already be known (even without invoking the SSD1306 template). Converting `DataTransmissionAdapter` to a template class did not yet work. WIP.

strongly-typed:
@salkinium This isn't as easy as expected.

Template parameter propagation does not work as I expect. Even making the Transmission a [template class](https://github.com/strongly-typed/xpcc/commit/39b257df733ab9fa99e6fc2b17429ffb66bf49a5) does not work.

Compiles with `BufferedGraphicDisplay<128, 64>`.
But does not compile with `BufferedGraphicDisplay<128, HHeight>`

I'm out of luck.

salkinium:
> Does buffer refer to buffer or to buffer?

That's messed up. Should be renamed to `data_buffer` and `display_buffer`. That may require some further refactoring in the code base though…

> I'm out of luck.

Whenever you're referring to a member of an inherited template class, you need to prefix it with `this->`. It has something to do with template instantiation time, didn't fully understand it.

Here are the tiny changes I did to make it work on top of your `fix/buffer` branch:
```patch
diff --git a/src/xpcc/driver/display/ssd1306.hpp b/src/xpcc/driver/display/ssd1306.hpp
index aace7d9..b26b513 100644
--- a/src/xpcc/driver/display/ssd1306.hpp
+++ b/src/xpcc/driver/display/ssd1306.hpp
@@ -150,9 +150,10 @@ public:
  * @ingroup    driver_display
  */
 template < class I2cMaster, uint8_t HHeight = 64 >
-class Ssd1306 : public ssd1306, public BufferedGraphicDisplay<128, 64>,
+class Ssd1306 : public ssd1306, public BufferedGraphicDisplay<128, HHeight>,
                                public I2cDevice<I2cMaster, 2, ssd1306::DataTransmissionAdapter<HHeight>>
 {
+       static_assert(HHeight == 64 or HHeight == 32, "Display height must be either 32 or 64 pixel!");
 public:
        Ssd1306(uint8_t address = 0x3C);

diff --git a/src/xpcc/driver/display/ssd1306_impl.hpp b/src/xpcc/driver/display/ssd1306_impl.hpp
index d036fc1..5dfc1c3 100644
--- a/src/xpcc/driver/display/ssd1306_impl.hpp
+++ b/src/xpcc/driver/display/ssd1306_impl.hpp
@@ -57,7 +57,7 @@ xpcc::Ssd1306<I2cMaster, HHeight>::startWriteDisplay()
 {
        RF_BEGIN();

-       RF_WAIT_UNTIL( this->transaction.configureDisplayWrite(pixbuf, 1024) and this->startTransaction() );
+       RF_WAIT_UNTIL( this->transaction.configureDisplayWrite(this->pixbuf, 1024) and this->startTransaction() );

        RF_END();
 }
```

strongly-typed:
Yeah! Love C++ Templates! This helps. 

Experimented with `BufferedGraphicDisplay<128, HHeight>::pixbuf` which is of course something completely different.

strongly-typed:
Just tested with [STM32L4 Nucleo](https://github.com/strongly-typed/xpcc/commits/feature/stm32l4_support_2017). Shall I stash `9ad3ae529995493f57514e1e1f25be9c58b9cb22` and `a2f64a6783c4b81ab3d26ac20285e279cb2e6afd`?

salkinium:
> Shall I stash

I think you meant squash. And yes, please.

strongly-typed:
Mash / Stash / Squash / whatever. Done.

salkinium:
> Mash / Stash / Squash / whatever. Done.

IT'S "WHATEVS". SHEESH.

salkinium (approved):

src/xpcc/ui/display/buffered_graphic_display.hpp
```diff
@@ -55,6 +55,10 @@ namespace xpcc
 		static_assert((Height % 8) == 0, "height must be a multiple of 8");
 
 	public:
+		static constexpr uint16_t display_buffer_width = Width;
+		static constexpr uint16_t display_buffer_height = Height / 8;
```
salkinium: Could you Capitalize this?
I find this super confusing especially in combination with `this->display_buffer_width`, which makes it look like a runtime variable.

salkinium (changes requested):

salkinium:
I don't actually like `BufferedGraphicDisplay` at all, it's a broken abstraction. The buffer really should be part of the driver, and `GraphicDisplay` writes into the display using a well defined interface.
This allows the driver to decide how many features to _implement_. For example display rotation can be expensive, or it could as simple as setting a config bit (the SSD1306 can do this for 180° rotation).
```
xpcc::Ssd1306<…> oled(…);
GraphicDisplay(oled) mainDisplay;
mainDisplay.rectangle(…);
```

salkinium (approved):

PR #228, opened 2017-03-19
[zmq] Make ZMQ Gateway example compile on OSX by default.

strongly-typed:
Using CanUSB instead of SocketCAN which is not available on OSX.

### DIFF
--- a/examples/zmq/2_zmq_gateway/main.cpp
+++ b/examples/zmq/2_zmq_gateway/main.cpp
@@ -4,7 +4,7 @@
 
 #include <xpcc/architecture/interface/can.hpp>
 #include <xpcc/architecture/platform/driver/can/canusb/canusb.hpp>
-#include <xpcc/architecture/platform/driver/can/socketcan/socketcan.hpp>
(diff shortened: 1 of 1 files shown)


salkinium:
Does this make the example work exclusively one or the other way?

strongly-typed:
Yes, it does. You need to switch between interfaces by commenting out one interface and enabling the other. See the second interface as free easter egg ;-). 

SocketCAN is not (yet) used by default by RCA. The CAN2USB is still connected by serial / USB interface. This may change as soon as the debug server is up and running again and an MCP2515 is hooked up by SPI to the debug server or the new [USB2CAN](https://github.com/roboterclubaachen/usb2can) is used.

Adding

     static xpcc::hosted::SocketCan canSocket;

in OSX will break because <strike>there is no SocketCAN implementation for OSX.</strike> [I did not yet look into SocketCAN driver rewritten from Linux to OS X](https://github.com/dn5/socketcanx) and [the popular Linux based can-utils and SocketCAN now available for OS X.](https://github.com/carloop/can-utils-osx)

salkinium:
Rebased and merged.

PR #229, opened 2017-03-22
[container] Improve BoundedDeque: Add element access operator and overwrite mode

mhthies:
The optional overwrite mode for `append()` and `prepend()` is nice to use the BoundedDeque as a ring buffer.

The bracket operator and the `get()` method can be used to access elements by id, e.g. to do a binary search on ordered elements in the deque.

### DIFF
--- a/src/xpcc/container/deque.hpp
+++ b/src/xpcc/container/deque.hpp
@@ -112,18 +112,84 @@ namespace xpcc
 		inline const T&
 		getFront() const;
 		
+		/**
+		 * \brief	Get item at specified index
+		 * 
+		 * Returns a reference to the item at index `n`, counting 0-indexed
+		 * from Front to Back, which is the same order in that the items were
+		 * appended and the iterator addresses them.
+		 * 
+		 * \warning Please make sure `n` is a valid index: 0 <= *n* < *size*.
+		 * 			Other indexes will cause undefined behaviour.
+		 */
+		inline T&
+		get(Index n);
+		
+		inline const T&
+		get(Index n) const;
+		
+		/**
+		 * \brief	Get item at specified index
+		 * 
+		 * Returns a reference to the item at index `n`, counting 0-indexed
+		 * from Front to Back.
+		 * 
+		 * \warning Please make sure `n` is a valid index: 0 <= *n* < *size*.
+		 * 			Other indexes will cause undefined behaviour.
+		 */
+		inline T&
+		operator[](Index n);
+		
+		inline const T&
+		operator[](Index n) const;
+		
+		/**
+		 * \brief	Get item at specified index
+		 * 
+		 * Returns a reference to the item at index `n`. The items are indexed
+		 * in reverse (Back to Front), which is their order when they have been
+		 * prepended.
+		 * 
+		 * \warning Please make sure `n` is a valid index: 0 <= *n* < *size*.
+		 * 			Other indexes will cause undefined behaviour.
+		 */
+		inline T&
+		rget(Index n);
+		
+		inline const T&
+		rget(Index n) const;
+		
 		inline T&
 		getBack();
 		
 		inline const T&
 		getBack() const;
 		
-		
 		bool
 		append(const T& value);
 		
+		/**
+		 * \brief	Append an item to the back of the deque overwriting existing items
+		 * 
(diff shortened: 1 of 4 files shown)


strongly-typed:
Very interesting!
What is the code size and runtime implications for that? It is probably OK, but I remember that this is used for processing of e.g. CAN messages in an Interrupt context. So it should be as lean as possible to avoid loosing CAN messages on high bus loads.

daniel-k:
> What is the code size and runtime implications for that? 

I didn't check code size, but runtime-wise it only introduces an additional bool test, which I guess should be fine.

If that's really a showstopper, we can optimize with templates (yay!):

```c++
#include <iostream>

template<typename T>
class Q
{
public:
    template<bool overwrite = false>
    void
    append(T item) {
        std::cout << "item: " << item << ", overwrite: " << overwrite << std::endl;
    }
};

int main()
{
    Q<int> queue;
    queue.append(1);
    queue.append<true>(2);
    return 0;
}
```

```
item: 1, overwrite: 0
item: 2, overwrite: 1
```

Doesn't break the interface (default parameter), but will duplicate the function when needed. So code size only increases when overwriting is used as I expect the compiler to optimize the dead-code in the default case away.

salkinium:
> used for processing of e.g. CAN messages in an Interrupt context. So it should be as lean as possible

It should be *reentrant*, which none of the containers really are. It may work for us now out of luck, because the functions are "lean", but the whole container stuff makes assumptions that may not be true on Cortex-M.

If we wanted to make them truly interrupt safe, we'd have to use [the atomic builtins of GCC](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html), which may efficiently map onto load/store exclusive on ARMv7-M and up and probably uses an atomic lock on AVR and ARMv6-M (not actually sure about that).

salkinium:
We are currently working on the next generation of this framework (over at [modm-io](https://github.com/modm-io), it's very much WIP). In the meantime we're forward porting all changes from xpcc to modm, which is licensed under MPLv2:

>
>
>

Since this is a non-trivial change to xpcc, **would you (@mhthies) be ok with licensing your contributions as MPLv2 as well?**

salkinium:
I like the intention and the code is fine, but I don't like the interface.
If I see it in the code, it'll look like this and I don't know what the `true` stands for:
```c++
queue.append(value, true);
queue.prepend(value, true);
queue.get(1, true);
```
What about?
```c++
queue.appendOverwrite(value);
queue.prependOverwrite(value);

queue.getReversed(1);
// maybe even rget?
queue.rget(1);
```

mhthies:
> Since this is a non-trivial change to xpcc, would you (@mhthies) be ok with licensing your contributions as MPLv2 as well?

Yep, of course. I'm currently not caring much about license details and I'm fine with any free software license. So MPLv2 seems okay as well.

> but I don't like the interface.
> If I see it in the code, it'll look like this and I don't know what the true stands for.

My intention was to modify the existing code and interface a minimum, but I see your point.

> What about?
>
> ```
> queue.appendOverwrite(value);
> queue.prependOverwrite(value);
> 
> queue.getReversed(1);
> // maybe even rget?
> queue.rget(1);
> ```

`queue.rget()` is probably nicer than my `reverse` parameter. Though, for `append()` and `prepend()`, I like the idea of having only one function for this task and disabling its overwrite-lock with an additional parameter. The non-template version would even allow to do this dynamically at runtime, although I can't think of a usecase for that.

salkinium:
> I like the idea of having only one function for this task and disabling its overwrite-lock with an additional parameter. The non-template version would even allow to do this dynamically at runtime, although I can't think of a usecase for that.

Hm, I'd still prefer the non-runtime version, particularly because you can still do the overwrite externally, if required:
```c++
if (overwrite) {
    queue.appendOverwrite(value);
} else {
    queue.append(value);
}
```
If `overwrite` is a template/constexpr parameter, the entire dead branch may be optimized away, which is probably harder to do if it's a function argument (even if it's a template class).

This optional implementation is also how the GPIO functions works, where you have a `set(void)`, `reset(void)`, but for convenience `set(bool)` which does the branching for you.

salkinium (approved):

salkinium:
Squashed and rebased. Many thanks for your contribution!

Issue #223, opened 2017-03-15
Flags, bitwise operators and conversion to flag groups 


chris-durand:
Instances of FlagsOperators, which are created when Flags are combined using bitwise operators, do not implicitly convert to a FlagsGroup.
However, FlagsOperators convert to Flags which themselves implicitly convert to FlagsGroup. Since the compiler will never apply more than one user-defined conversion automatically, you need an explicit cast.

Adding a FlagsOperators conversion constructor to the FlagsGroup class template should eliminate unnecessary casts without allowing additional undesired conversions.
[Here](https://github.com/chris-durand/xpcc/commit/df1df54e6a0014e3d58a9dbc60b625f1728e4837) you can find an implementation.

Did I miss some unintended side effects?

salkinium:
.oO( Le sigh, C++ is such a terrible language. )

I remember adding FlagsOperators rather hastily, so it's very likely I missed this.
The best way to make sure this does what it's supposed to do is to [add a unittest](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/architecture/interface/test/register_test.cpp) that fails before this fix is applied, and passes afterwards.

chris-durand:
> The best way to make sure this does what it's supposed to do is to add a unittest that fails before this fix is applied, and passes afterwards.

I have no idea how to write a reasonable unittest for this feature. This is a language issue during compilation. Before the fix the code would simply refuse to compile instead of failing at runtime.

salkinium:
Well, do you have an example of code that compiles _with_ this change, but doesn't _without_ it? That's your unittest.

chris-durand:
> Well, do you have an example of code that compiles with this change, but doesn't without it? That's your unittest.

I'll add the test soon.

> .oO( Le sigh, C++ is such a terrible language. )

It will get even better... I encountered another issue after adding the suggested conversion operators. In some driver code I defined two function overloads of which one takes a FlagsGroup and the other one takes a uint8_t. Everything works as intended unless you dare to pass bitwise or-ed flags to it. The compiler complains about the overload being ambiguous although it is clear the flags don't implicitly convert to uint8_t.

Here is where the fun part starts. All Flag* classes inherit from Register which defines explicitly deleted int conversion operators to prevent unintentional implicit casts to int types through the bool conversion operator. The overload resolution fails because the compiler sees the deleted uint8_t conversion operator.

[...]

PR #230, opened 2017-03-24
[architecture] Fix implicit conversion of Flags

chris-durand:
Fix of the issues discussed in #223. Conversion of Flags to an associated FlagsGroup will now work correctly even when multiple enum constants or flag variables are combined with logical operators.

The bool conversion of flags classes is implemented in terms of `explicit operator bool()` to fix issues with function overloading on Flags parameters.

The explicit operator will only allow implicit casts where contextual conversion takes place, which is more restrictive than the previous implementation. It will occur in the following contexts:
- conditions of if, while, for, do-while statements
- logical operators (&&, ||)
- negation (operator !)
- static_assert

This may break user code that requires implicit bool conversions on other occasions, for instance in return statements but I don't expect a lot of people to rely on that. Outside of the flag implementation itself I only changed 5 lines of code in xpcc for this reason. 
 
I had a look at the code that is not build in examples and unittests but uses flags. These changes shouldn't do any harm there but I wouldn't bet my life on that.

### DIFF
--- a/src/unittest/harness.hpp
+++ b/src/unittest/harness.hpp
@@ -262,7 +262,7 @@ namespace unittest
 }
 
 #define	TEST_ASSERT_TRUE(expr)	\
-	TEST_RETURN__(::unittest::checkExpression((expr), __LINE__))
+	TEST_RETURN__(::unittest::checkExpression(static_cast<bool>(expr), __LINE__))
 
 #define	TEST_ASSERT_FALSE(expr)	\
 	TEST_RETURN__(::unittest::checkExpression(!static_cast<bool>(expr), __LINE__))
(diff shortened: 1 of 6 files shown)


salkinium (approved):

salkinium:
Thanks for getting to the bottom of this, this is a much nicer solution than using `= delete` for everything.

> I don't expect a lot of people to rely on that. Outside of the flag implementation itself I only changed 5 lines of code in xpcc for this reason.

I suspect the RCA code may break at some points, so keep an eye out for that.

chris-durand:
> I suspect the RCA code may break at some points, so keep an eye out for that.

I've just built most of our current repos with these changes. Everything compiled without errors.

PR #234, opened 2017-03-28
Driver and example for VL53L0X distance sensor 

chris-durand:
A driver for ST's VL53L0X distance sensor. Since ST does not provide any useful documentation how to use the device, this implementation models the behaviour of the ST VL53L0X API. Don't blame me for the magic values, blame ST ;-).


salkinium:
I'll give this a more thorough review soon, but at first glance this is very beautiful code :-)

examples/nucleo_f401re/distance_vl53l0/main.cpp
```diff
@@ -0,0 +1,158 @@
...
+			PT_WAIT_UNTIL(this->timeout.isExpired());
+		}
+
+		XPCC_LOG_DEBUG << "Device initialized" << xpcc::endl;
+
+		// set measurement time to 200ms (high accuracy mode)
+		if(not PT_CALL(distance.setMaxMeasurementTime(MeasurementTimePrecision)))
+			XPCC_LOG_DEBUG << "Setting measurement time failed" << xpcc::endl;
+		else
+			this->highAccuracyMode = true;
```
salkinium: Braces please

src/xpcc/driver/position/vl53l0.hpp
```diff
@@ -0,0 +1,531 @@
...
+	};
+
+public:
+	struct xpcc_packed
+	Data
+	{
+		template< class I2cMaster >
+		friend class Vl53l0;
+
+		static constexpr uint16_t InvalidDistance = 8190;
```
salkinium: I'd be more comfortable if this were written as `0x1FFE`, cos then the binary AND in
```
if((distance & InvalidDistance) == InvalidDistance) {
```
makes more sense.

src/xpcc/driver/position/vl53l0.hpp
```diff
@@ -0,0 +1,531 @@
...
+		RangeErrorCode error;
+	};
+}; // struct vl53l0
+
+/**
+ * VL53L0X: Proximity sensor module
+ *
+ * The class is build for single-shot measurements.
+ *
+ * @ingroup driver_position
```
salkinium: Add yourself as `@author`.

src/xpcc/driver/position/vl53l0.hpp
```diff
@@ -0,0 +1,531 @@
...
+	/// @endcond
+
+	/// @cond
+	/// periodically read an 8 bit value from a register until the supplied
+	/// predicate returns true
+	///
+	/// the predicate must take a uint8_t argument and return a bool
+	template < typename Predicate >
+	xpcc::ResumableResult<bool>
+	poll(Register reg, Predicate pred, uint16_t timeoutMs = 20, uint16_t stepMs = 4);
```
salkinium, line 432: Am I missing something? Shouldn't the `Predicate` just be a C-style function pointer?
You can declare it inline too:
```cpp
poll(Register reg, bool (*pred)(uint8_t), ...);
```

src/xpcc/driver/position/vl53l0_impl.hpp
```diff
@@ -0,0 +1,633 @@
+// coding: utf-8
+/* Copyright (c) 2017, Christopher Durand, Niklas Hauser
```
salkinium: Nah, just you. This is plenty original work, not really a derivative of mine.

src/xpcc/driver/position/vl53l0_impl.hpp
```diff
@@ -0,0 +1,633 @@
...
+	constexpr MSRCConfig_t msrcConfig = MSRCConfig::DisableRateCheck
+		| MSRCConfig::DisablePreRangeCheck;
+
+	RF_BEGIN();
+
+	// read the measurement sequence configuration into this->sequenceInfo
+	VL53L0_RF_CALL(readSequenceInfo());
+
+	// "Set I2C standard mode"
+	VL53L0_RF_CALL(write(Register::UNDOCUMENTED__I2C_MODE, 0x00));
```
salkinium, line 61: _blames ST_

src/xpcc/driver/position/vl53l0.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+namespace xpcc
+{
+
+namespace vl53l0_private
+{
+
+FLASH_STORAGE(BinaryConfiguration configurationFlash[]) =
+{
+	// the infamous binary blob of black magic
```
salkinium, line 20: _blames ST_

examples/nucleo_f401re/distance_vl53l0/main.cpp
```diff
@@ -0,0 +1,158 @@
+#include <xpcc/architecture/platform.hpp>
```
salkinium, line 10: copyright missing.

src/xpcc/driver/position/vl53l0.cpp
```diff
@@ -0,0 +1,106 @@
+// coding: utf-8
+/* Copyright (c) 2017, Christopher Durand, Niklas Hauser
```
salkinium: just you.

src/xpcc/driver/position/vl53l0_impl.hpp
```diff
@@ -0,0 +1,633 @@
...
+	VL53L0_RF_CALL(write(Register(0xFF), 0x00));
+	VL53L0_RF_CALL(write(Register(0x80), 0x00));
+
+	// start range measurement
+	VL53L0_RF_CALL(write(Register::SYSRANGE__START, Start::StartStop));
+
+	// wait for the start flag to be cleared, use default timeout
+	VL53L0_RF_CALL(poll(Register::SYSRANGE__START, [](uint8_t value) {
+		return not (value & uint8_t(Start::StartStop));
+	}));
```
salkinium, line 160: 👏 for use of lambda. This is very, very nice.

src/xpcc/driver/position/vl53l0_impl.hpp
```diff
@@ -0,0 +1,633 @@
...
+
+	// wait for the measurement to finish, timeout of 2 * this->measurementTimeUs
+	VL53L0_RF_CALL(poll(Register::RESULT__INTERRUPT_STATUS, [](uint8_t value) {
+		return value & (InterruptStatus::NewSampleReady | InterruptStatus::OutOfWindow).value;
+	}, measurementTimeUs / 1000 * 2));
+
+	// read 16 bit range value
+	VL53L0_RF_CALL(read(Register::RESULT__RANGE_VALUE0, &i2cBuffer[1], 2));
+	data.distance = (i2cBuffer[1] << 8) + i2cBuffer[2];
```
salkinium: See comment about `uint16_t distance`.

src/xpcc/driver/position/vl53l0_impl.hpp
```diff
@@ -0,0 +1,633 @@
...
+	VL53L0_RF_CALL(write(Register::DYNAMIC_SPAD__REF_EN_START_OFFSET, 0x00));
+	VL53L0_RF_CALL(write(Register::DYNAMIC_SPAD__NUM_REQUESTED_REF_SPAD, 0x2C));
+	VL53L0_RF_CALL(write(Register(0xFF), 0x00));
+	VL53L0_RF_CALL(write(Register::GLOBAL__CONFIG_REF_EN_START_SELECT, 0xB4));
+
+	if(not setupReferenceSpadMap(spadInfo.map , &i2cBuffer[1])) {
+		XPCC_LOG_ERROR << "Invalid SPAD data in non-volatile memory.\n";
+		XPCC_LOG_ERROR << "A full 'SPAD management' recalibration has to be performed.\n";
+		XPCC_LOG_ERROR << "This procedure is not implemented in this driver.\n";
+		XPCC_LOG_ERROR << "Please use the VL53L0X API provided by ST.\n" << xpcc::endl;
```
salkinium: double newline

src/xpcc/driver/position/vl53l0_impl.hpp
```diff
@@ -0,0 +1,633 @@
...
+	} else if(sequenceInfo.enabledSteps & MeasurementSequenceStep::MSRC) {
+		const uint16_t stepTime = TimeOverhead::MSRC + msrcDssTccTimeout;
+		if(finalRangeTime <= stepTime) {
+			return false;
+		}
+
+		finalRangeTime -= stepTime;
+	}
+
+	if(sequenceInfo.enabledSteps & MeasurementSequenceStep::PreRange) {
```
salkinium: here and above, for multi-liners please put the `{` on the next line. [See source file conventions](https://github.com/roboterclubaachen/xpcc/blob/develop/doc/coding_convention.md#examples).

src/xpcc/driver/position/vl53l0_impl.hpp
```diff
@@ -0,0 +1,633 @@
...
+xpcc::Vl53l0<I2cMaster>::encodeTimeout(uint32_t timeoutMclks)
+{
+	// Decompose timeoutMclks into the following expression:
+	// coefficient*(2^exponent) + 1
+
+	uint16_t encodedValue = 0;
+	uint32_t coefficient = 0;
+	uint16_t exponent = 0;
+
+	if (timeoutMclks > 0) {
```
salkinium: Sometimes it's `if (` other times it's `if(`. I don't mind, I just find it funny. Also `{` on newline pls.

src/xpcc/driver/position/vl53l0_impl.hpp
```diff
@@ -0,0 +1,633 @@
...
+	index = timeoutMs;
+
+	while(true) {
+		if ( not RF_CALL(read(reg, i2cBuffer[1])) ) {
+			break;
+		}
+
+		if(pred(i2cBuffer[1])) {
+			RF_RETURN(true);
+		}
```
salkinium, line 625: 👏

src/xpcc/driver/position/vl53l0.hpp
```diff
@@ -0,0 +1,531 @@
...
+			}
+
+			return (error == RangeErrorCode::RangeComplete
+				|| error == RangeErrorCode::NoError);
+		}
+		
+		inline RangeErrorCode
+		getRangeError() { return error; }
+	private:
+		uint16_t distance;
```
salkinium: Hm, this may break endianness (in theory).

The idea behind making a Data class and instantiating it separately is that it would allow us to place several data objects into a packed struct and just send that sequentially over a network.
This means that the endianness of the data object must the the same on all devices, and the easiest way to do that is just to keep it the way it falls out of the sensor.
I would therefore prefer, if you just stored the raw bytes in the Data class and did the endianness conversion in the `Data::getDistance()`.

But of course this is all BS, since every platform we support is little-endian, incl. every Intel OS, so this is a bit academic. I'd be fine if you left it like this.

salkinium (changes requested):
Very nice code for such a complex sensor!
Still a few minor things to consider.

src/xpcc/driver/position/vl53l0.hpp
chris-durand: The C style function pointer won't allow you to use a lambda with a capture. It's more explicit but less flexible.

ekiwi: > won't allow you to use a lambda with a capture

I assume because the closure needs to be stored somewhere and a C pointer just stores a pointer, correct?

chris-durand: I suppose so.

The standard says:
> The closure type for a lambda-expression with **no lambda-capture** has a public non-virtual non-explicit const conversion function to pointer to function having the same parameter and return types as the closure type’s function call operator.

src/xpcc/driver/position/vl53l0_impl.hpp
chris-durand: I probably started with `if(`, copied code from the Vl6180 driver which has `if (` and got finally confused by that. I'll unify the formatting.

src/xpcc/driver/position/vl53l0.hpp
chris-durand: Ah, now this data struct makes more sense to me. I just followed the Vl6180 code for the sake of consistency.

src/xpcc/driver/position/vl53l0.cpp
chris-durand: This line can also be found in the Vl6180 driver (vl6180.cpp). It might be appropriate to change it there.

salkinium: Nono, leave it. We are blaming ST.

chris-durand:
Done.

src/xpcc/driver/position/vl53l0_impl.hpp
```diff
@@ -0,0 +1,641 @@
...
+		return not (value & uint8_t(Start::StartStop));
+	}));
+
+	// wait for the measurement to finish, timeout of 2 * this->measurementTimeUs
+	VL53L0_RF_CALL(poll(Register::RESULT__INTERRUPT_STATUS, [](uint8_t value) {
+		return value & (InterruptStatus::NewSampleReady | InterruptStatus::OutOfWindow).value;
+	}, measurementTimeUs / 1000 * 2));
+
+	// read 16 bit range value
+	VL53L0_RF_CALL(read(Register::RESULT__RANGE_VALUE0, &data.distanceBuffer[0], 2));
```
salkinium: Careful, this can lead to data "tearing" issues, since the I2C operates asynchronously. If you call `getData().isValid()` from a protothread while running this function from another, you may catch the case where just the first byte is updated which may give you false positives.

Using the `i2cBuffer` and then copying it over "proto-atomically" to the distance buffer will avoid that. All other sensors use this technique as well.

src/xpcc/driver/position/vl53l0_impl.hpp
```diff
@@ -0,0 +1,641 @@
...
+	registerValue = encodeTimeout(finalRangeTime);
+	return true;
+}
+
+template < typename I2cMaster >
+uint32_t
+xpcc::Vl53l0<I2cMaster>::timeoutToMicroseconds(uint16_t mclks, uint8_t vcselPeriodPclks)
+{
+	// in nanoseconds
+	const uint32_t macroPeriodNs = ((UINT32_C(2304) * vcselPeriodPclks * 1655) + 500) / 1000;
```
salkinium, line 481: Why not use `uint32_t(2304)` here and below? I don't think you're outsmarting the compiler here ;-P

salkinium (changes requested):

src/xpcc/driver/position/vl53l0_impl.hpp
chris-durand: Strictly speaking thats not a cast but the platform independent version of declaring a uint32_t literal. This is defined in the cstdint header. See [this](http://en.cppreference.com/w/cpp/types/integer) for more details. I can use the cast if you prefer that.

src/xpcc/driver/position/vl53l0_impl.hpp
chris-durand: You are right. Change is on the way.

src/xpcc/driver/position/vl53l0_impl.hpp
salkinium: No, let's leave it like this.

salkinium (approved):

PR #235, opened 2017-04-01
add --specs=nosys.specs to linker flags

salkinium:
Not entirely certain if this is the right thing to do, do lets see what the CI says.

### DIFF
--- a/scons/site_tools/arm.py
+++ b/scons/site_tools/arm.py
@@ -208,6 +208,7 @@ def generate(env, **kw):
 		"-Wl,-wrap,_calloc_r",
 		"-Wl,-wrap,_realloc_r",
 		"-Wl,-wrap,_free_r",
+		"--specs=nosys.specs",
 		"--specs=nano.specs",
 		"-nostartfiles",
 		"-L$LINKPATH",


salkinium:
Builds just fine, confirmed correct linking behaviour manually with F469 examples.

scons/site_tools/arm.py
```diff
@@ -208,6 +208,7 @@ def generate(env, **kw):
 		"-Wl,-wrap,_calloc_r",
 		"-Wl,-wrap,_realloc_r",
 		"-Wl,-wrap,_free_r",
+		"--specs=nosys.specs",
 		"--specs=nano.specs",
```
ekiwi, line 212: what does gcc do with two specs files?
is one ignored or are they merged somehow?

salkinium: They are merged explicitly. This is done in the specs directly, see the sources for [`nano.specs`](https://sourceware.org/git/gitweb.cgi?p=newlib-cygwin.git;a=blob;f=libgloss/arm/elf-nano.specs;h=82594bd03cea8584188f625826deb20a50fee603;hb=HEAD#l12) and [`nosys.specs`](https://sourceware.org/git/gitweb.cgi?p=newlib-cygwin.git;a=blob;f=libgloss/libnosys/nosys.specs;h=3a84db3b53299ac769cac645cf437aefcaa68071;hb=HEAD#l7). Interesting how there is a mini "build system" in gcc too.

ekiwi: omg...

PR #241, opened 2017-04-10
Add .mailmap file for correct author assignment

salkinium:
GitHub apparently [does not use the `.mailmap`](https://github.com/holman/ama/issues/551) for assigning authorship in the stats. But maybe it changed since 2014.

@daniel-k: which email do you prefer?
@dergraaf: is this ok for you?

cc @strongly-typed @ekiwi

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,29 @@
+Antal Szabó <szabo.antal.92@gmail.com>
+Christian Menard <christian.menard@gmx.de>
+Christoph Rüdi <christoph.ruedi@rwth-aachen.de>
+Christopher Durand <christopher.durand@rwth-aachen.de>
+Daniel Krebs <github@daniel-krebs.net>
+Daniel Krebs <github@daniel-krebs.net> <da.krebs@web.de>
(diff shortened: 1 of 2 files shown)


daniel-k:
I'm currently only using `github@daniel-krebs.net`

dergraaf:
> @dergraaf: is this ok for you?

Fine by me.
Is there any reason why the file contains the same mail address twice for some people?

salkinium:
> Is there any reason why the file contains the same mail address twice for some people?

Yes, these people have committed via the same `user.email`, but different `user.name`s. See [the `.mailmap` documentation](https://git-scm.com/docs/git-shortlog#_mapping_authors).

dergraaf:
> Yes, these people have committed via the same user.email, but different user.names. See the .mailmap documentation.

But wouldn't you need only one address in that case (like for Martin E.)? For Martin R. and Thorsten you use the same address twice. From my understanding that is redundant.

salkinium:
> From my understanding that is redundant.

Yep, you're right! This works without the duplicate email. Coooool…

PR #244, opened 2017-04-24
Feature/tcs3471driver

arjunsarin:
old TCS3414 color sensor driver modified for new TCS3471X color sensor.


salkinium:
Code has been tested in hardware. Both sensor drivers could use a rewrite, but then again we've never documented how to do sensor drivers "our" way, so whatever ;-)

salkinium:
Merged manually, thanks!

PR #249, opened 2017-05-09
[stm32l4] Simple support of ADC in STM32L4.

strongly-typed:
Tested with Nucleo-L476RG.

Interstingly the ADC of STM32F3 and STM32L4 is quite similar.


salkinium:
Also update [the DFG please](https://github.com/roboterclubaachen/xpcc/blob/develop/tools/device_file_generator/dfg/stm32/stm_writer.py#L57), otherwise I may forget when regenerating.

salkinium (changes requested):
- Put the changes to gpio.hpp into one commit.
- Only update the DFG, then regen.

salkinium (approved):

PR #251, opened 2017-05-10
Driver for DS1302 and some goodies.

strongly-typed:


salkinium:
I like your `examples/generic` idea a lot! We can also get rid of a lot of `{board}/blink` examples by adding a way to specify the `build.board` via the command line and building this folder with multiple boards via TravisCI.

Furthermore I'm thinking of exporting the entire folder to `getting-started-with-xpcc` with every release to get good and tested examples. I've been hesitant to have to maintain two repos, if we can automate the manual sync away, the overall experience become much better for everyone.

src/xpcc/architecture/platform/board/nucleo64_arduino.hpp
```diff
@@ -0,0 +1,30 @@
+// Nucleo64 Arduino Header Footprint
+
+#ifndef XPCC_STM32_NUCLEO_ARDUINO_HPP
+#define XPCC_STM32_NUCLEO_ARDUINO_HPP
```
salkinium:
```diff
- XPCC_STM32_NUCLEO_ARDUINO_HPP
+ XPCC_STM32_NUCLEO64_ARDUINO_HPP
```

src/xpcc/architecture/platform/board/nucleo64_arduino.hpp
```diff
@@ -0,0 +1,30 @@
...
+using D6  = GpioB10;
+using D7  = GpioA8;
+using D8  = GpioA9;
+using D9  = GpioC7;
+using D10 = GpioB6;
+using D11 = GpioA7;
+using D12 = GpioA6;
+using D13 = GpioA5;
+using D14 = GpioB9;
+using D15 = GpioB8;
```
salkinium, line 28: I was hesitant to do such a thing, since there is one Nucleo64 (F334) whose pinout differs in 3 pins. But lets solve it when it becomes an issue.

src/xpcc/driver/rtc/ds1302/ds1302.hpp
```diff
@@ -0,0 +1,153 @@
...
+		getYear() {
+			return data[6];
+		}
+
+		inline bool
+		isWriteProtected() {
+			return data[7] & (1 << 7);
+		}
+
+		///
```
salkinium: yes?

src/xpcc/driver/rtc/ds1302/ds1302.hpp
```diff
@@ -0,0 +1,153 @@
...
+ *
+ * @ingroup  ds1302
+ * @author   Sascha Schade (strongly-typed)
+ */
+template < class PinSet >
+class Ds1302
+{
+	using Ce   = typename PinSet::Ce;
+	using Sclk = typename PinSet::Sclk;
+	using Io   = typename PinSet::Io;
```
salkinium, line 113: This is a good idea, but I think an inherited base struct would be clearer from a documentation view (see https://github.com/roboterclubaachen/xpcc/pull/86#issuecomment-186578136 for demo).  Also the class should probably be called `Config` for conformance with future drivers using this for more than just pins.

salkinium (changes requested):
Documentation of the driver config struct needs some investigation, rest is ok.
Still amused that the real time clock driver has a bus interface that can only be bit banged, lol.

src/xpcc/architecture/platform/board/nucleo64_arduino.hpp
salkinium: Hey, one more project name and I have Bingo! BWAHAHA

strongly-typed:
I adopted your proposal and changed it to an inherited `Config` class.

salkinium:
> I adopted your proposal and changed it to an inherited Config class.

Oh, we'll bastardize C++ in The Name of Embedded™ till Stroustrup begs for mercy. BWHAHAHA!

salkinium (approved):

PR #259, opened 2017-06-30
[travis] Update avr-gcc to v5.4.0.

salkinium:
This refactors the travis dependency installation to use pre-compiled binaries for all toolchains and for boost. The cache is used to store the compressed packages only, which reduces cache size.

I also modified our example runner to allow compiling of multiple example folders in one go. This allows to utilize the test instances much better, offsetting the initial setup costs and ordering test instances to run in better order to minimize global running time. Good example sets and order was determined experimentally.
This reduces TravisCI running time to <20mins down from 35-40mins.


salkinium:
And ironically the CI fails on this test, cos I merged it too quickly. Rest assured: it passed.

PR #260, opened 2017-06-30
[travis] Do not use caching.

salkinium:
Testing if it's faster without caching.


salkinium:
It's not faster without caching, if caching works right (most of the time).
Downloading the cache is about 3s vs. 60-100s from the original hosts.

PR #261, opened 2017-06-30
[tools] Add Python 3 support to xml parser

mhthies:
… doing the smallest possible changes to the code and keeping compatibility to python 2.7.

This is required to use the xml parser in new Python projects, that extract their information about the xpcc project directly from the xml files. I did not check the builder (code generator) for Python 3 support yet.

Unfortunately custom compare operators for objects get a little trickier in Python 3 and require some new checks. See https://docs.python.org/3.0/whatsnew/3.0.html#ordering-comparisons

### DIFF
--- a/tools/system_design/xmlparser/action.py
+++ b/tools/system_design/xmlparser/action.py
@@ -1,10 +1,10 @@
 #!/usr/bin/env python2
 # -*- coding: utf-8 -*-
 
-import utils
-import xml_utils
+from . import utils
+from . import xml_utils
 
-from parser_exception import ParserException
+from .parser_exception import ParserException
 
 class Action(object):
 	
--- a/tools/system_design/xmlparser/action.py
+++ b/tools/system_design/xmlparser/action.py
@@ -31,10 +31,20 @@ class Action(object):
 				raise ParserException("Type '%s' is not defined. Used by Action '%s')" % (type, self.name))
 				
 		return type
-	
+
 	def __cmp__(self, other):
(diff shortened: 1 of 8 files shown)


salkinium:
>  I did not check the builder (code generator) for Python 3 support yet.

Ok, holding off merging then until tested.

mhthies:
> I did not check the builder (code generator) for Python 3 support yet.

This is not required, since we currently use Python 2 for code generation and my changes to the xml parser don't break Python 2 support (which I tested with RCA robot code). External Python 3 projects probably only need the xml parser, not the code builder, so I didn't touch the builder yet.

salkinium:
Ok, asking @dergraaf for a review, since I don't know the code that well.

dergraaf:
> This is not required, since we currently use Python 2 for code generation 

Would it be possible to change that to Python 3 as well? Keeping compatibility between Python 2 and 3 will be annoying in the long run.

mhthies:
> Would it be possible to change that to Python 3 as well?

That should be possible with some changes, similar to the commits above. But I didn't want to change the Python interpreter of the xpcc build tools in the moment, as this would add a new dependency [1] and need more thorough testing. If we still decide to change to Python 3, we should also port all the other python scripts. 

I just need the xpcc XML parser for a Python 3 project at the RCA, so I patched it to be compatible with Python 3, without breaking the xpcc build system.

[1] We would force users to have Python 3, but Scons still requires Python 2: http://scons.org/faq.html#Am_I_restricted_to_using_Python_2.4_code_in_my_SConscript_files.3F

PR #263, opened 2017-07-04
[io] Check arguments of printf format string

Maju-Ketchup:
XPCC is missing checking the arguments of printf
When using `XPCC_LOG_INFO.printf("%d", (double)66.0)` a "0"  is written into the log stream w/out compile time warning

https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#Common-Function-Attributes

### DIFF
--- a/src/xpcc/io/iostream.hpp
+++ b/src/xpcc/io/iostream.hpp
@@ -460,10 +460,10 @@ public:
 	 * @param	fmt		Format string
 	 */
 	IOStream&
-	printf(const char* fmt, ...);
+	printf(const char* fmt, ...)  __attribute__((format(printf, 2, 3)));
 
 	IOStream&
-	vprintf(const char *fmt, va_list vlist);
+	vprintf(const char *fmt, va_list vlist) __attribute__((format(printf, 2, 0)));
 
 protected:
 	void


src/xpcc/io/iostream.hpp
```diff
@@ -460,7 +460,7 @@ class IOStream
 	 * @param	fmt		Format string
 	 */
 	IOStream&
-	printf(const char* fmt, ...);
+	printf(const char* fmt, ...)  __attribute__((format(printf, 2, 3)));
 
 	IOStream&
 	vprintf(const char *fmt, va_list vlist);
```
salkinium: Please add it to this function too.

salkinium (changes requested):
Very nice, such an improvement.

src/xpcc/io/iostream.hpp
Maju-Ketchup: If I've read the GCC doc correctly you cannot apply the attribute check to vprintf like functions the only thing you can do is checking the string for correctness. 
`For functions where the arguments are not available to be checked (such as vprintf), specify the third parameter as zero. In this case the compiler only checks the format string for consistency.` -gcc documentation
But I assume that this is better than nothing and will add this in a free minute.

Maju-Ketchup: -Done-

salkinium (approved):
Thanks!

Maju-Ketchup:
Compiling the Linux example will now throw some Warnings of people who used the printf function wrongly before:
```
xpcc/build/linux/printf/libxpcc/generated_platform/driver/core/hosted/assert.cpp:71:78: warning: format ‘%d’ expects argument of type ‘int’, but argument 4 has type ‘uintptr_t {aka long unsigned int}’ [-Wformat=]
  if (context) { XPCC_LOG_ERROR.printf(" @ %p (%d)", (void *) context, context); }
```
and:
```
xpcc/src/xpcc/ui/gui/widgets/numberfield.cpp:57:33: warning: format ‘%ld’ expects argument of type ‘long int’, but argument 3 has type ‘int’ [-Wformat=]
  out->printf("%03ld", afterComma);
```

strongly-typed:
Good spot! There are more warnings of that kind:

* [can_message.cpp](https://travis-ci.org/roboterclubaachen/xpcc/jobs/250345555#L1269)
* And many in [bme280_data_impl_fp.hpp](https://travis-ci.org/roboterclubaachen/xpcc/jobs/250345566#L1241) and related files.

daniel-k:
> `xpcc/src/xpcc/ui/gui/widgets/numberfield.cpp:57:33: warning: format ‘%ld’ expects argument of type ‘long int’, but argument 3 has type ‘int’ [-Wformat=] out->printf("%03ld", afterComma);`

Well that was probably me 😄 This &^%& gui code becomes slightly embarassing in retrospect :D

strongly-typed:
> `can-message.cpp`
> `bme280_data_impl_fp.hpp`

Well, that was probably me 😄 This &^%& test  code becomes slightly embarassing in retrospect :D

salkinium:
Time to build ~a~ with `-Wall`.

PR #272, opened 2017-07-27
[math] filter: fix moving average for negative averages

daniel-k:
Since template parameter N is defined as `std::size_t` which is unsigned, the result of the average calculation will be implicitly casted and therefore negative averages give wrong results.

I figure `int` is the right type to replace since it's the same size on the systems that I'm aware of.


daniel-k:
I just found out that there's a unittest that should even cover this case. However, with my fix it fails ... Reading a relevant [StackOverflow answer](https://stackoverflow.com/a/3602857) I've got the impression, that the test data is wrong.

```
FAIL: moving_average_test:120 : -1 == -2
FAIL: moving_average_test:120 : -2 == -3
```

Test data:
```c
static const TestData data[] =
	{
		{ 10,	2 },
		{ 10,	5 },
		{ 10,	7 },
		{ 10,	10 },
		{ -10,	5 },
		{ -10,	0 },
		{ 5,	-2 },
		{ 5,	-3 },
		// ...
	};
```

Doing the calculations by hand, the average for line `{ 5,	-2 }` should be `-1.25`. Since the standard says that the fractional part will be discarded, the correct (integer) average would `-1`. Same goes for the other offending line.

chris-durand:
> I figure int is the right type to replace since it's the same size on the systems that I'm aware of.

Yes and no. You are right for all microcontroller targets supported in xpcc, but on x86-64 Linux int is 32 bits and size_t has the size of a pointer which is 64 bit. On 64 bit Windows even long has a size of 32 bits.

daniel-k:
Oh, you're right! Until now, I assumed that would be true for 64bit Linux too. However, as long as the replacement type can store 65535, everything is fine for our implementation, since this is the limit `uint_fast16_t`.

ekiwi:
> I figure int is the right type to replace since it's the same size on the systems that I'm aware of.

`N` is of type `std::size`, because it is the _size_ of the buffer used by the filter. Imho, the correct thing to do would be to cast it to `T` before performing any calculations: `std::static_cast<T>(N)`

salkinium:
Any update on this? cc @dergraaf

ekiwi:
> Any update on this?

I would like @daniel-k to address my comment above.

dergraaf:
I agree with @ekiwi and @daniel-k. `N` should indeed stay as `size_t` and should only be cast when used. On the other hand the test data is wrong and should be changed as @daniel-k indicated. With both changes the unittests pass again.

Not sure why the tests work at the moment.

daniel-k:
Updated and rebased on current develop. Thanks for the comments!

dergraaf (approved):

PR #275, opened 2017-08-11
Driver and example for AMSYS 5915 pressure sensor

rleh:
Driver for the [AMSYS 5915](http://www.amsys.de/sheets/amsys.de.ams5915.pdf) pressure sensor series.
Based on the HCLAx pressure sensor driver.

Tested in hardware (AMS 5915-0050-D).

### DIFF
--- a/src/xpcc/driver/pressure/amsys5915.hpp
+++ b/src/xpcc/driver/pressure/amsys5915.hpp
@@ -1,0 +1,131 @@
+// coding: utf-8
+/* Copyright (c) 2012, Niklas Hauser
+ * Copyright (c) 2017, Raphael Lehmann
+ * All Rights Reserved.
+ *
+ * The file is part of the xpcc library and is released under the 3-clause BSD
+ * license. See the file `LICENSE` for the full license governing this code.
+ */
+// ----------------------------------------------------------------------------
+
+#ifndef XPCC_AMSYS5915_HPP
+#define XPCC_AMSYS5915_HPP
+
+#include <xpcc/architecture/interface/i2c_device.hpp>
+#include <xpcc/math/utils/endianness.hpp>
+
+namespace xpcc
+{
+
+// forward declaration for friending with amsys5915::Data
+template < typename I2cMaster >
+class Amsys5915;
+
+struct amsys5915
+{
+	struct xpcc_packed
+	Data
+	{
+		template < typename I2cMaster >
+		friend class Amsys5915;
+
+	public:
+		/**
+		 * This method returns the pressure as a normalized float from 0-1.
+		 * You have to scale and offset this according to the specific sensor
+		 * you have.
+		 * So if you have AMS 5915-0010-D, you can measure +10mBar, you need to
+		 * multiply it with 10.0f to get the pressure in mBar.
+		 * If you have AMS 5915-0350-D-B, which can measure ±350mBar you first need to
+		 * subtract 0.5f and then multiply it with 350.0f!
+ 		 */
+		float
+		getPressure()
+		{
+			// mask undefined bits
+			data[0] &= 0b00111111;
+			// Full scale span is 13107, with offset 1638
+			uint16_t *rData = reinterpret_cast<uint16_t*>(data);
+			uint16_t pressure = xpcc::fromBigEndian(*rData) - 1638;
+			return static_cast<float>(pressure) / 13107.f;
+		}
+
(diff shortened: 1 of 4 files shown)


src/xpcc/driver/pressure/amsys5915.hpp
```diff
@@ -0,0 +1,133 @@
...
+ 		 */
+		float
+		getPressure()
+		{
+			// mask undefined bits
+			data[0] &= 0b00111111;
+			// Full scale span is 13107, with offset 1638
+			uint16_t *rData = reinterpret_cast<uint16_t*>(data);
+			uint16_t pressure = xpcc::fromBigEndian(*rData) - 1638;
+			return float(pressure) / 13107;
```
ekiwi: `static_cast<float>(pressure) / 13107.f`

src/xpcc/driver/pressure/amsys5915.hpp
```diff
@@ -0,0 +1,133 @@
...
+		}
+
+		/**
+		 * This method returns the temperature of the pressure sensor in °C.
+		 */
+		float
+		getTemperature()
+		{
+			uint16_t temperatureRaw = (data[3] >> 5) | (data[2] << 3);
```
ekiwi: is `data[2]` implicitly cast to `uint16_t` here?

ekiwi: (before the shift?)

src/xpcc/driver/pressure/amsys5915.hpp
```diff
@@ -0,0 +1,133 @@
...
+		RF_END_RETURN( this->wasTransactionSuccessful() );
+	}
+
+	/// reads the Pressure registers and buffers the results
+	xpcc::ResumableResult<bool>
+	readPressure()
+	{
+		RF_BEGIN();
+
+		RF_END_RETURN_CALL( this->runTransaction() );
```
salkinium: just do `return this->runTransaction();` and inline the function.

src/xpcc/driver/pressure/amsys5915.hpp
rleh: [Copy&Paste](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/driver/pressure/hclax.hpp#L47)

src/xpcc/driver/pressure/amsys5915.hpp
dergraaf: It is implicitly converted to `int` which is fine for 16 bit operations. For 32 bit an explicit cast is necessary for architectures where `int` is 16 bit.

src/xpcc/driver/pressure/amsys5915.hpp
```diff
@@ -0,0 +1,133 @@
...
+ * @author	Raphael Lehman, Niklas Hauser
+ */
+template < typename I2cMaster >
+class Amsys5915 : public amsys5915, public xpcc::I2cDevice<I2cMaster, 1, I2cReadTransaction>
+{
+public:
+	/**
+	 * @param	data	a amsys5915::Data object
+	 * @bug The address of the sensor is by factory default set to 0x28.
+	 *      This means you cannot use two HCLA sensors on the same bus!
```
salkinium: HCLA?

salkinium (changes requested):
Cosmetic changes, otherwise a-ok.

src/xpcc/driver/pressure/amsys5915.hpp
ekiwi: Thanks for the clarification.

I was wondering why the following code did not exert undefined behavior:
```.cpp
uint8_t data = 255;
uint16_t shifted_data_0 = data << 8;
uint16_t shifted_data_1 = data << 16;
```

Since `data` is implicitly casted to `int` which is 32bit on my machine, you need to shift by `32` for it to be undefined behavior:

```.cpp
uint16_t shifted_data_2 = data << 32;
```

leads to:
```
runtime error: shift exponent 32 is too large for 32-bit type 'int'`
```

src/xpcc/driver/pressure/amsys5915.hpp
rleh: Again copy&paste 😀 
I'll fix it

salkinium (dismissed):

PR #278, opened 2017-08-17
Driver: MCP2515: Revive driver and add 8 MHz config.

strongly-typed:
Tested in hardware with 8 MHz external crystal for
10, 20, 50, 100, 125, 250, 500 and 1000 kBps with
STM32 bxCAN and oscilloscope with protocol decoder.


src/xpcc/driver/can/mcp2515_impl.hpp
```diff
@@ -92,9 +90,11 @@ xpcc::Mcp2515<SPI, CS, INT>::initialize(uint32_t bitrate)
 	writeRegister(BFPCTRL, 0);
 
 	// check if we could read back some of the values
-	if (readRegister(CNF2) != cfgPtr[static_cast<uint8_t>(bitrate) * 3 + 1])
+	uint8_t readback = readRegister(CNF2);
+	if (readback != cnf[CNF2_idx])
 	{
 		// we are not able to communicate with the MCP2515 => abort
+		XPCC_LOG_ERROR.printf("MCP2515 readback failed. Received: %02x, expected %02x\n", readback, cnf[CNF2_idx]);
```
salkinium: Also add `xpcc_assert` here!

salkinium (changes requested):
Nice

strongly-typed:
[Like so?](https://github.com/strongly-typed/xpcc/commit/5c93979000ab81a8abaf536793a5a2da3415f7da)

src/xpcc/driver/can/mcp2515_impl.hpp
```diff
@@ -423,3 +437,17 @@ xpcc::Mcp2515<SPI, CS, INT>::readIdentifier(uint32_t& identifier)
...
+    	!strncmp(location, "init",     4) and
+    	!strncmp(failure,  "readback", 8))
+    {
+        return xpcc::Abandonment::Fail;
+    }
+    return xpcc::Abandonment::DontCare;
+}
+
+// Register assertion handler with system
+XPCC_ASSERTION_HANDLER(can_assertion_handler);
```
salkinium: No, the user defines the handlers, not the driver.

salkinium (changes requested):
The driver only calls the `xpcc_assert`, it does not declare the handler. Also see ["The Curious Case of xpcc's Error Model"](http://blog.salkinium.com/xpccs-error-model). Here is an [example of multiple assertion handlers](https://github.com/roboterclubaachen/xpcc/blob/develop/examples/stm32f469_discovery/assert/main.cpp) being used.

src/xpcc/driver/can/mcp2515_impl.hpp
strongly-typed: And when he/she forgets to define one? How does she/he know about the necessity to define one? Oh, I see

> There is no way of knowing if the function you’re calling can fail an assert [...]

Hmm, only a mother could love that?

Any chance to get a weak default handler?

salkinium: > And when he/she forgets to define one?

The default behavior is to abandon the program execution. This is the de-facto behavior when failing an `assert`. However, this abandonment can be overwritten by the user via an assertion handler.

> How does she/he know about the necessity to define one?

Through documentation. There should be a description of assertions somewhere™, probably in the function's doxygen. The idea was that you could scan the source code for `xpcc_assert` and the containing function and autogenerate this documentation. But I didn't get to that yet.

> Any chance to get a weak default handler?

I'm confused about what you want (presumably aggregated by poor documentation on my part). The assertion handler above _is_ a "strong default handler", cos you `Abandonment::Fail` if this assertion is triggered! If you wanted a "weak" one, you would have to return `Abandonment::Ignore` (which is overruled by any assertion handler returing `Abandonment::Fail`).

But why would you want to do that? Your previous commit with the `XPCC_LOG_ERROR` made it seem like this is an irrecoverable error, hence the `xpcc_assert`. If this is not the case, and you can recover from the error, return an error code.
There's also `xpcc_assert_debug()` which is enabled in debug builds (which we don't have, so it's currently never enabled ;-). 

Both `xpcc_assert` and `xpcc_assert_debug` return the condition, so you can do stuff like this:
```cpp
if (not xpcc_assert(test_condition, ...))
{
    // user chose to ignore assertion
    // lets log this failure, but continue anyways
    XPCC_LOG_ERROR << "foobar failed, system degraded, implosion imminent" << xpcc::endl;
}
```

> Hmm, only a mother could love that?

Yes, of course. _And_ I wrote an extensive blog post about the research and implementation… double bonus 😛

examples/avr/can/mcp2515/main.cpp
```diff
@@ -32,6 +32,20 @@ FLASH_STORAGE(uint8_t canFilter[]) =
 	MCP2515_FILTER_EXTENDED(0),	// Mask 1
 };
 
+xpcc::Abandonment can_assertion_handler(const char *module, const char *location, const char *failure, uintptr_t)
+{
+    if (!strncmp(module,   "mcp2515",  7) and
+    	!strncmp(location, "init",     4) and
+    	!strncmp(failure,  "readback", 8))
+    {
+        return xpcc::Abandonment::Fail;
```
salkinium: Abandonment _is_ the default behavior, you don't need to explicitly code it.

examples/avr/can/mcp2515/main.cpp
```diff
@@ -32,6 +32,20 @@ FLASH_STORAGE(uint8_t canFilter[]) =
...
+    }
+    return xpcc::Abandonment::DontCare;
+}
+
+// Register assertion handler with system
+XPCC_ASSERTION_HANDLER(can_assertion_handler);
```
salkinium: Assertion handlers are for _ignoring_ error, ie. whitelisting them. What you did is blacklist your assertion, it can never be ignored by anyone anymore. Remove this atrocity at once, dear sir! _\*sips tea aggressively\*_

salkinium (changes requested):

examples/avr/can/mcp2515/main.cpp
salkinium: The `xpcc_assert` by default behaves like a normal `assert`, it abandons program execution if it fails. You don't need to do anything special to enable this behavior. 

Only if you want to ignore an assertion, then you can define an assertion handler that returns `xpcc::Abandonment::Ignore` when it matches the assertion identifiers.

strongly-typed:
OK, gotcha.

But the user has to define the `xpcc_abandon` handler. Which btw is missing for Nucleo L476RG board.

     ➜  platform git:(feature/stm32f107) ✗ ack HAS_LOGGER | grep board
     board/arduino_uno/arduino_uno.hpp:19:#define XPCC_BOARD_HAS_LOGGER
     board/nucleo_f031k6/nucleo_f031k6.hpp:19:#define XPCC_BOARD_HAS_LOGGER
     board/nucleo_f103rb/nucleo_f103rb.hpp:19:#define XPCC_BOARD_HAS_LOGGER
     board/nucleo_f303k8/nucleo_f303k8.hpp:18:#define XPCC_BOARD_HAS_LOGGER
     board/nucleo_f401re/nucleo_f401re.hpp:19:#define XPCC_BOARD_HAS_LOGGER
     board/nucleo_f411re/nucleo_f411re.hpp:18:#define XPCC_BOARD_HAS_LOGGER
     board/nucleo_f429zi/nucleo_f429zi.hpp:18:#define XPCC_BOARD_HAS_LOGGER
     board/nucleo_l476rg/nucleo_l476rg.hpp:21:#define XPCC_BOARD_HAS_LOGGER
     board/stm32f469_discovery/stm32f469_discovery.hpp:23:#define XPCC_BOARD_HAS_LOGGER
     board/stm32f746g_discovery/stm32f746g_discovery.hpp:21:#define XPCC_BOARD_HAS_LOGGER
     board/stm32f769i_discovery/stm32f769i_discovery.hpp:21:#define XPCC_BOARD_HAS_LOGGER
     ➜  platform git:(feature/stm32f107) ✗ ack xpcc_abandon | grep xpcc_abandon
     board/nucleo_f031k6/nucleo_f031k6.cpp:20:xpcc_abandon(const char * module,
     board/nucleo_f103rb/nucleo_f103rb.cpp:20:xpcc_abandon(const char * module,
     board/nucleo_f303k8/nucleo_f303k8.cpp:20:xpcc_abandon(const char * module,
     board/nucleo_f401re/nucleo_f401re.cpp:20:xpcc_abandon(const char * module,
     board/nucleo_f411re/nucleo_f411re.cpp:20:xpcc_abandon(const char * module,
     board/nucleo_f429zi/nucleo_f429zi.cpp:20:xpcc_abandon(const char * module,
     board/stm32f469_discovery/stm32f469_discovery.cpp:49:xpcc_abandon(const char * module,
     board/stm32f746g_discovery/stm32f746g_discovery.cpp:29:xpcc_abandon(const char *      module,
     board/stm32f769i_discovery/stm32f769i_discovery.cpp:28:xpcc_abandon(const char *      module,

salkinium:
> But the user has to define the `xpcc_abandon` handler.

Yes, it's weakly defined to be empty by default, since a logger may not be available on very small devices.

> Which btw is missing for Nucleo L476RG board.

Want to add it copy-paste from another STM32 board? They are all the same.

salkinium (approved):

PR #283, opened 2017-08-22
MCP2515: Fix: Do not acknowledge unread data.

### LINKED ISSUE #282: MCP2515: Dropping packets by acknowledging unread data
In [MCP2515 driver](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/driver/can/mcp2515_impl.hpp#L227) the receive flags are cleared twice. Once by rising `CS`, and once by explicitly by clearing the flag using SPI command. 

The `READ RX BUFFER Instruction` clears this flag automatically. There is no need to clear it manually (MCP2515 datasheet, revision H, section 12.4, page 65).

> by automatically clearing the associated receive flag, RXnIF (CANINTF), when CS is raised at the end of the command.

If another packet is received between rise of `CS` and clearing the flag manually, it will be dropped. 

Package loss was proven by placing a short delay in between and firing CAN messages to the MCP2515.

strongly-typed:
This fixes #282

### DIFF
--- a/src/xpcc/driver/can/mcp2515_impl.hpp
+++ b/src/xpcc/driver/can/mcp2515_impl.hpp
@@ -224,13 +224,8 @@ xpcc::Mcp2515<SPI, CS, INT>::getMessage(can::Message& message)
 	}
 	chipSelect.set();
 
-	// clear interrupt flag
-	if (status & FLAG_RXB0_FULL) {
-		bitModify(CANINTF, RX0IF, 0);
-	}
-	else {
-		bitModify(CANINTF, RX1IF, 0);
-	}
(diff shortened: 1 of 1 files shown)


salkinium (dismissed):

salkinium:
Race conditions are always such fun.

strongly-typed:
I actually did not run into it 😌. Discovered by reading the datasheet.

I had to provoke it manually by adding a small `xpcc::delayMicroseconds` in between. With a high CAN baud rate and a slow processor / slow [SPI implementation](xpcc.io) it could have happened in the real world. I need soon talk about xpcc's SPI implementation.

PR #285, opened 2017-08-24
[scons] Add DFU programming method to universal SConstruct.

strongly-typed:
Maybe that is a bit broad? Did not find an easy check for STM32 in scons.

### DIFF
--- a/scons/SConstruct
+++ b/scons/SConstruct
@@ -89,6 +89,10 @@ else:
         env.Alias('debug', env.GdbDebug(program))
         env.Alias('bin', env.Bin(program))
 
+        # Most STM32's can be programmed by USB DFU if board allows
+        binfile = env.Bin(program)
+        env.Alias('dfu', env.DfuStm32Programmer(binfile))
+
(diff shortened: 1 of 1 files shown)


salkinium:
So the way the programming method is currently chosen is [by checking what the user put in the `project.cfg` file](https://github.com/roboterclubaachen/xpcc/blob/develop/scons/site_tools/xpcc.py#L527-L556). I'd add a check for `[dfu]` and perhaps parse sub configs if required.

strongly-typed:
See the USB2CAN based on an DFU-capable STM32. It can be either programmed by SWD (three test pads for that) or by bringing the STM32 into DFU mode by holding down the DFU button. So it is not neither nor, it is **AND**.  RCA's default was `scons program` (OpenOCD) and `scons dfu` (DFU).

Both programming methods have benefits:
* DFU: No ST Link available and working hard- and software.
* ST Link: Problems with USB and/or want to debug.

salkinium (approved):
🚢🕴

PR #292, opened 2017-09-23
[travis] Use SCons v2.5.1.

salkinium:
SCons recently released v3.0.0 which supports Python 3.5+, which however
breaks our build system for now.

**THIS IS A QUICKFIX!**

We need to upgrade our build system for Python 3.5+, otherwise new developers who install SCons v3 will have a bad initial experience!


salkinium:
Upgrade tracked in #293.

PR #298, opened 2017-10-11
[ci] Enable CircleCI to compile our code in Docker containers.

strongly-typed:
Please do not merge before status badge is correctly displayed. This is still a bit wip.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,216 @@
+version: 2
+jobs:
+  "hosted-unittests":
+    docker:
+      - image: roboterclubaachen/xpcc-build:latest
+    environment:
+      - SCONS_LIB_DIR: "/usr/local/lib/python2.7/dist-packages/scons-2.5.1"
+      - SCONSFLAGS: "-j3"
+    steps:
+      - checkout
+      - run:
+          name: Check environment
+          command: |
+            python --version  || true
+            python2 --version || true
+            python3 --version || true
+            which scons
+            scons --version
+            which avr-g++
+            avr-g++ --version
+            which arm-none-eabi-g++
+            arm-none-eabi-g++ --version
+
+      - run:
+          name: Unittest hosted
+          command: |
+            scons unittest
+
+  "stm32-unittests":
+    docker:
+      - image: roboterclubaachen/xpcc-build:latest
+    environment:
+      - SCONS_LIB_DIR: "/usr/local/lib/python2.7/dist-packages/scons-2.5.1"
+      - SCONSFLAGS: "-j3"
+    steps:
+      - checkout
+      - run:
+          name: Unittest STM32
+          command: |
+            scons unittest target=stm32
+
+  "atmega-unittests":
+    docker:
+      - image: roboterclubaachen/xpcc-build:latest
+    environment:
+      - SCONS_LIB_DIR: "/usr/local/lib/python2.7/dist-packages/scons-2.5.1"
+      - SCONSFLAGS: "-j3"
+    steps:
+      - checkout
+      - run:
+          name: Unittest AVR ATmega
+          command: |
+            scons unittest target=atmega
+
+  "stm32f0-examples":
+    docker:
+      - image: roboterclubaachen/xpcc-build:latest
+    environment:
+      - SCONS_LIB_DIR: "/usr/local/lib/python2.7/dist-packages/scons-2.5.1"
+      - SCONSFLAGS: "-j3"
+    steps:
+      - checkout
+      - run:
+          name: Examples STM32F0 Series
+          command: |
+            scons check=examples examples=stm32f0_discovery,stm32f072_discovery,nucleo_f031k6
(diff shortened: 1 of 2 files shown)


strongly-typed:
First bug:

screen shot 2017-10-11 at 16 58 33

It should take care about the workflow, not a single job.

strongly-typed:
"ci/circleci: ut-stm32 — Waiting for status to be reported"

There seems to be some ["Degraded Performance"](https://status.circleci.com) issues with CircleCI at the moment. The test succeeded. Maybe the status will be reported later.

strongly-typed:
Retriggering the build correctly updated the CircleCI statuses.

salkinium:
Nice, `life.fifo.push(review);`.

.circleci/config.yml
```diff
@@ -0,0 +1,216 @@
+version: 2
+jobs:
+  "ut-hosted":
+    docker:
+      - image: stronglytyped/xpcc-build-arm:latest
```
salkinium: Do we want to move this docker image to this account? Maybe with a fork? Would like to have this self contained.

.circleci/config.yml
```diff
@@ -0,0 +1,216 @@
...
+      - SCONS_LIB_DIR: "/usr/local/lib/python2.7/dist-packages/scons-2.5.1"
+      - SCONSFLAGS: "-j3"
+    steps:
+      - checkout
+      - run:
+          name: Examples LPCxpresso Board
+          command: |
+            scons check=examples examples=lpcxpresso
+
+  "ex-linux":
```
salkinium: Why `ex-` and not `example-`? Is it a char limitation somewhere?

.circleci/config.yml
```diff
@@ -0,0 +1,216 @@
...
+      - "ut-stm32"
+      - "ut-atmega"
+      - "ex-stm32f0"
+      - "ex-stm32f1"
+      - "ex-stm32f3"
+      - "ex-stm32f7"
+      - "ex-stm32l4"
+      - "ex-lpcxpresso"
+      - "ex-linux"
+      - "ex-avr"
```
salkinium: I like this split very much more than the monoblock from travis, makes it much easier to see the failure on GitHub.

salkinium (commented):
I’d prefer to have this account fork your docker container, then use that. The fork preserves your authorship, if you’re into that. It just seems more canonical to have everything xpcc related under this account. What do you think?

.circleci/config.yml
strongly-typed: I would think this is a feature of CircleCI. I have no idea how to split Travis into multiple parts.

.circleci/config.yml
strongly-typed: Yes, in the workflow view of CircleCI
screen shot 2017-10-12 at 00 51 40

.circleci/config.yml
strongly-typed: Probably, but I would like to rename it to something like `xpcc-build` before, getting rid of the `arm` because I recently added the AVR toolchain.

strongly-typed: BTW, it's like a Matryoshka doll:
```
stronglytyped/xpcc-build-arm <- stronglytyped/arm-none-eabi-gcc <- ubuntu:16.04
```

.circleci/config.yml
salkinium: Would it be acceptable to switch it? E.g. `stm32f4-example`, `atmega-unittest`?

.circleci/config.yml
salkinium: Ok, I _think_ you can fork to this account, ping me if you can’t.

strongly-typed: I just did. Wondering if there is already a hub.docker.com account associated with RCA? 

Private and strictly-confidential: ;-) Whose email address is used for the github.com/roboterclubaachen account? Should the same address be used for hub.docker.com account?

salkinium: @RoboterClubAachen is an organization and as such must be owned by at least one account. It therefore doesn't have its own email. Nonetheless, it *does* have a billing email, which is our org email (see [our profile info](https://github.com/organizations/roboterclubaachen/settings/profile)). 

You should create an [organiziation on Docker Hub](https://docs.docker.com/docker-hub/orgs/) so that ownership isn't tied to a specific account for life (analog reasoning with this GitHub org).

salkinium: Oh, and you've been upgraded to owner. Happy responsibiliting…

strongly-typed: Just found the organization on Docker Hub and just created us. ;-)

strongly-typed: Needed some fiddling to get the accounts linked. Finally, I deleted everything and reconnected it.

salkinium (commented):
👍 this is a very nice addition.

Maybe squash commit 2 & 4 into 1, then you could fast forward merge too.

PR #304, opened 2017-10-25
[stm32] Implement SpiHal::enable() for Spi 4, 5 and 6

chris-durand:
`SpiHal{{ id }}::enable()` was only implemented for Spi1, Spi2 and Spi3 …
Tested in Hardware for Spi5 on STM32F429ZI.


salkinium (approved):

salkinium (changes requested):
You also need to implement this in `disable()`.

chris-durand:
Fixed

salkinium (approved):
🚢🕴

PR #306, opened 2017-10-26
Add rosserial communication to xpcc

strongly-typed:
xpcc has a great hardware support for sensors and different microcontroller platforms.
ROS is a great framework for dealing with robots and rosserial is the natural way to bridge between ROS and any microcontroller. Unfortunately rosserial only supports Arduino and not xpcc.

Adding this PR would allow ROS users to use xpcc (instead of Arduino) on the microcontroller and transfer data to and from ROS (e.g. environmental data from a sensor connected to a microcontroller.)

This PR will also allow xpcc to talk to any ROS enabled robot, e.g. publishing a ROS `geometry_msgs/Twist` message to a `/cmd_vel` topic on a Turtlebot3 Burger for rapid prototyping or receive odometry updates from a laser scan matcher running ROS.

Generated files from rosserial are added as a subtree in `ext/ros_lib`.

I'm open for any discussion, modification and demonstration of this code.


salkinium (dismissed):
LGTM. I don't like that the rosserial code is compiled for every example, even those not using it, but that's a separate build system problem and out-of-scope here.

salkinium:
@strongly-typed Feel free to rebase and merge these commits with a nice merge message (similar to the one above).

PR #309, opened 2017-11-02
[avr] Fix compilation for avr gcc 7.2.0

### LINKED ISSUE #308: AVR target is broken?
I've found that compiling for AVR target (arduino uno board) fails with following diagnostics:
`xpcc/src/xpcc/math/geometry/location_2d_impl.hpp:169:51: error: call of overloaded 'abs(float)' is ambiguous
    (std::abs(this->orientation - other.orientation) < __FLT_EPSILON__));`

Reproduction is very simple: clone getting-started-with-xpcc repo, change board in project.cfg to arduino_uno and then run scons.

chris-durand:
Fixes #308

### DIFF
--- a/src/xpcc/ui/display/graphic_display.cpp
+++ b/src/xpcc/ui/display/graphic_display.cpp
@@ -28,7 +28,7 @@
  */
 // ----------------------------------------------------------------------------
 
-#include <stdlib.h>
+#include <cstdlib>
 #include <xpcc/math/utils/bit_operation.hpp>
 
 #include "font/fixed_width_5x8.hpp"


siy:
Perhaps it worth to get rid of stdlib.h completely. There are other files which include this header file.  I guess that under some circumstances they may trigger similar issue.

salkinium (approved):

salkinium:
There are quite a few uses of [`stdlib.h`](https://github.com/roboterclubaachen/xpcc/search?p=3&q=stdlib.h&type=&utf8=%E2%9C%93)

siy:
Can /src/xpcc/ui/display/buffered_graphic_display.hpp cause such an issue? It also includes stdlib.h.

chris-durand:
> Perhaps it worth to get rid of stdlib.h completely.

The better solution would be to fix the underlying cause of that problem. I am able to reproduce this issue in a simple example that does not use any xpcc code except for the cstdlib, cmath, … headers xpcc provides for avr. It seems this compile error occurs if and only if all three headers `cmath`, `cstdlib` and `stdlib.h` are included and `stdlib.h`, but not `cstdlib`, is included before `cmath` 😒 .

This works (compiled with `avr-g++ test.cpp -Ixpcc/src/stdc++/`):
```
#include <stdlib.h>
#include <cstdlib>
#include <cmath>

int main()
{
        std::abs(1.0f);
}
```
This fails with `call of overloaded 'abs(float)' is ambiguous`:
```
#include <stdlib.h>
#include <cmath>
#include <cstdlib>

int main()
{
        std::abs(1.0f);
}
```

I'd suggest to merge this PR as a quick fix. I don't have time to look into this now.

salkinium:
We try to provide our [own subset of the std C++ lib for AVRs](https://github.com/roboterclubaachen/xpcc/blob/develop/src/stdc%2B%2B/cstdlib#L108-L118). @dergraaf added this once a long time ago in a repo far far away.

chris-durand:
I know. In the test examples above I added these files to the include path to test them separately from the rest of xpcc. We clearly have a problem with these headers when the correct compilation depends on the order of includes … Maybe it's an avr-gcc bug.

salkinium:
> Maybe it's an avr-gcc bug.

Dunno, could be a @dergraaf bug, I mean, when was the last time we ran unit-tests on him?!? 🙈

PR #311, opened 2017-11-03
fix for static serial interface, and added a example

cajt:
This fixes some minor bugs in the static uart wrapper, and adds an example for testing.

### DIFF
--- a/examples/linux/static_serial_interface/SConstruct
+++ b/examples/linux/static_serial_interface/SConstruct
@@ -1,0 +1,4 @@
+# path to the xpcc root directory
+xpccpath = '../../..'
+# execute the common SConstruct file
+execfile(xpccpath + '/scons/SConstruct')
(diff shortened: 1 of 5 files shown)


salkinium (approved):
Nice, thanks!

salkinium:
@cajt  as you've probably know, we are currently working on the next generation of this framework (over at [modm-io](https://github.com/modm-io)), which will be licensed under MPLv2 (see below for why). Since this is a non-trivial change to xpcc, would you be ok with licensing your current and future contributions as MPLv2 as well?

> 
>

cajt:
On 06.11.2017 00:42, Niklas Hauser wrote:
>
> @cajt <https://github.com/cajt> as you've probably know, we are
> currently working on the next generation of this framework (over at
> modm-io <https://github.com/modm-io>), which will be licensed under
> MPLv2 (see below for why). Since this is a non-trivial change to xpcc,
> would you be ok with licensing your current and future contributions
> as MPLv2 as well?
>
That step is fine with me, go ahead.

Greetings, Calle

-- 
Carl Johann Treudler
Mobile +49 177 190 66 89

PR #312, opened 2017-11-04
fix non-blocking writes with linux's rfcomm (bluetooth)

cajt:
Fix for handling non-blocking writes to Linux's rfcomm devices. 
Prior code reported errors and dropped the character. This behaviour seems to be triggered by the rfcomm device in non-blocking mode (doesn't happen if used with usb-serial devices).

### DIFF
--- a/src/xpcc/architecture/platform/driver/uart/posix/serial_interface.cpp
+++ b/src/xpcc/architecture/platform/driver/uart/posix/serial_interface.cpp
@@ -271,9 +271,21 @@ xpcc::hosted::SerialInterface::write(char c)
 /*	SUB_LOGGER_LOG(logger, Logger::ERROR, "writeByte")
 		<< "0x" << std::hex << (int)data << "; ";
  */
-	int reply = ::write(this->fileDescriptor, &c, 1);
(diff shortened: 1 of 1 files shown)


src/xpcc/architecture/platform/driver/uart/posix/serial_interface.cpp
```diff
@@ -271,9 +271,19 @@ xpcc::hosted::SerialInterface::write(char c)
...
+	// retry on EAGAIN, otherwise report error
+	// some linux rfcomm devices need this
+	while(1) {
+		reply = ::write(this->fileDescriptor, &c, 1);
+		if (reply <= 0) {
+			if (errno!=EAGAIN) {
+				this->dumpErrorMessage();
+				break;
+			}
+  	} else {
```
salkinium: Oh, this indentation is weird, I thought it was a while-else construct.

salkinium (changes requested):
Please fix indentation. Otherwise 👍

salkinium:
Please rebase and squash into one commit on top of develop.

PR #313, opened 2017-11-05
[architecture] Allow using XPCC_FLAGS macros in template classes

chris-durand:
Declaring flags inside a template class using the `XPCC_FLAGS8`/`XPCC_FLAGS16`/`XPCC_FLAGS32` macros fails to compile. `Parent::EnumType` in lines 923 and 930 of register.hpp may be a dependent name, thus we have to add `typename` here.


salkinium (dismissed):
Nice!

salkinium:
Do you want to add this to our unit test? Then it won’t be broken again on refactoring.

chris-durand:
I will add this to the unit test tomorrow.

chris-durand:
Oh, I missed the function body. More `typename` needed …

salkinium (approved):
Thanks!

PR #315, opened 2017-11-17
[fix][io]fixed that only low 32Bit of an uint64_t were shown in Logstream

Maju-Ketchup:
While working with 40+Bit Register I stored into an uint64_t, I wanted to put them via printf into my  debug Logstream.
Problem was only the low 32Bits were shown in the log.

`//Example code`
`uint64_t a = 0xFEDCBA9876543210;`
`XPCC_LOG_DEBUG.printf("%llx",a);`
`// Output is "76543210"`

We (@AndreGilerson and I) fixxed the issue in the iostream.hpp and  iostream_printf.cpp editing the "long" variables to "long long"

### DIFF
--- a/src/xpcc/io/iostream.hpp
+++ b/src/xpcc/io/iostream.hpp
@@ -300,7 +300,7 @@ public:
 
 // The 64-bit types on the AVR are extremely slow and are
 // therefore excluded here
-#if !defined(XPCC__CPU_AVR)
+#if not defined(XPCC__CPU_AVR)
 	xpcc_always_inline IOStream&
 	operator << (const uint64_t& v)
 	{
@@ -502,13 +502,18 @@ protected:
 	void
 	writeFloat(const float& value);
 
-#if !defined(XPCC__CPU_AVR)
+#if not defined(XPCC__CPU_AVR)
 	void
 	writeDouble(const double& value);
 #endif
 
-	void
-	writeUnsignedInteger(unsigned long unsignedValue, uint_fast8_t base, size_t width, char fill, bool isNegative);
+        void
+        writeUnsignedInteger(unsigned long unsignedValue, uint_fast8_t base, size_t width, char fill, bool isNegative);
+#if not defined(XPCC__CPU_AVR)
+        void
+        writeUnsignedLongLong(unsigned long long unsignedValue, uint_fast8_t base, size_t width, char fill, bool isNegative);
+#endif
+
 
 private:
 	enum class
--- a/src/xpcc/io/iostream_printf.cpp
+++ b/src/xpcc/io/iostream_printf.cpp
@@ -167,6 +167,21 @@ xpcc::IOStream::vprintf(const char *fmt, va_list ap)
 			// Print fractional part
 			writeUnsignedInteger((unsigned int)float_value, base, width_frac, '0', false);
 		}
+#if not defined(XPCC__CPU_AVR)
+		else if (isLongLong)
+		{
+			long long signedValue = va_arg(ap, long long);
+			if (isSigned)
+			{
+				if (signedValue < 0)
+				{
+					isNegative = true;
+				    signedValue = -signedValue; // make it positive
+				}
+			}
+			writeUnsignedLongLong((unsigned long long) signedValue, base, width, fill, isNegative);
+		}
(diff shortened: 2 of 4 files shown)


salkinium:
1. Well… that commit message definitely needs to be fixxed 😜
2. Can you please add a unit test for 64-bit integers?
3. The use of 64-bit integers is going to hurt AVR performance quite a bit. Please create a typedef for the `long long`s you replaced and point them to `long` for `XPCC__CPU_AVR`. We'll deal with enabling 64-bit support on AVRs later.

Maju-Ketchup:
I'm on it.
I'll try to implement the unit tests.
And yes, I am "**fixing**" the commit message -.-' (stupid English language with its unpredictable exceptions)

Maju-Ketchup:
Done.

salkinium (dismissed):
Thanks!

strongly-typed:
@Maju-Ketchup 
Could you please assess memory and performance impacts? Yes, `printf` is a heavy piece of cake and xpcc's `printf` is a (false?) compromise between functionality and size/speed for microcontrollers. It's always unclear how to optimise. As 32 bit is the native data width for Cortex-M, I would focus on that. For that `uint64_t` corner case some other solution might be appropriate.

Maju-Ketchup:
An other alternative would be an extra function only for long longs this won't hurt the performance of the other types, but this would make it possible to display 64bit values as well.

salkinium:
What about spitting the 64bit into two 32bit prints?  The logic for this would go into the [printf string parsing function](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/io/iostream_printf.cpp#L178), not the formatting function. This wouldn't slow down every integer print.

Maju-Ketchup:
My thougths  - I'll do it tomorrow morning.

ekiwi:
How about just using [IO Streams](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/io/iostream.hpp#L305)?
Then you do not have to deal with all the varg function weirdness in C/C++.

ekiwi:
We could also investigate a more C++ way of implementing `printf`.
Something like: https://github.com/c42f/tinyformat

Maju-Ketchup:
Using printf has the precious ability to display numbers to a diffrent base (eg 2 or 16) which is pretty awesome while debuging registers. Tiny format looks intresting and I will take a look into it.

chris-durand:
> We could also investigate a more C++ way of implementing printf.

👍. Variadic templates are a nice way to implement a type safe printf, but you have to keep in mind the code bloat this might cause. For every combination of arguments and in every translation unit the compiler will instantiate the function template. I don't know if this is a practical problem on AVR.

Template functions are a possibility to avoid the overhead of uint64_t operations when printing smaller types. The arithmetic operations could be moved to a template function taking the required int type as a template argument.

ekiwi:
> Using printf has the precious ability to display numbers to a diffrent base (eg 2 or 16)

https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/io/iostream.hpp#L107
https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/io/iostream.hpp#L115

Maju-Ketchup:
Made a different function to print those 64bit values.
I just wanted this fixed for now and leaving future changes / ideas to you right now.

src/xpcc/io/iostream.hpp
```diff
@@ -507,8 +507,13 @@ class IOStream
 	writeDouble(const double& value);
 #endif
 
-	void
-	writeUnsignedInteger(unsigned long unsignedValue, uint_fast8_t base, size_t width, char fill, bool isNegative);
+        void
+        writeUnsignedInteger(unsigned long unsignedValue, uint_fast8_t base, size_t width, char fill, bool isNegative);
+#if !defined(XPCC__CPU_AVR)
```
strongly-typed: `if not defined`
Please be consistent to in regards to line 170.

strongly-typed (changes requested):

Maju-Ketchup:
@strongly-typed corrections done.

strongly-typed:
Merged manually.

Issue #85, opened 2016-02-09
Matrix/Vector Multiplications


ekiwi:
I failed multiple times to multiply a `xpcc::Matrix4f` with a `xpcc::Vector4f`.

``` .cpp
const xpcc::Vector4f v0;
const xpcc::Matrix4f m0;
const auto v1 = m0 * v0;
```

fails with the following error:

```
main.cpp: In function ‘int main()’:
main.cpp:46:21: error: ambiguous overload for ‘operator*’ (operand types are ‘const Matrix4f {aka const xpcc::Matrix<float, 4u, 4u>}’ and ‘const Vector4f {aka const xpcc::Vector<float, 4u>}’)
  const auto v1 = m0 * v0;
                     ^
/.../xpcc/src/xpcc/math/geometry/vector4_impl.hpp:826:1: note: candidate: xpcc::Vector<T, 4u> operator*(const xpcc::Matrix<T, 4u, 4u>&, const xpcc::Vector<T, 4u>&) [with T = float; U = float]
 operator * (const xpcc::Matrix<T, 4, 4> &lhs, const xpcc::Vector<U,4> &rhs)
 ^
/.../xpcc/src/xpcc/math/geometry/vector4.hpp:269:28: note: candidate: xpcc::Vector<T, 4u> xpcc::operator*(const xpcc::Matrix<T, 4u, 4u>&, const xpcc::Vector<T, 4u>&) [with T = float; U = float]
  static inline Vector<U,4> operator * (const Matrix<T, 4, 4> &lhs, const Vector<U,4> &rhs)
                            ^
/.../xpcc/src/xpcc/math/geometry/vector4.hpp:263:28: note: candidate: xpcc::Vector<T, 4u> xpcc::operator*(const U&, const xpcc::Vector<T, 4u>&) [with U = xpcc::Matrix<float, 4u, 4u>; T = float]
  static inline Vector<T,4> operator * (const U &lhs, const Vector<T,4> &rhs)
```

Using `asTransposedMatrix` which I found in [vector4.hpp](https://github.com/roboterclubaachen/xpcc/blob/9a5213e4aa06819b2e1eeb7e423039b1ecbac0ba/src/xpcc/math/geometry/vector4.hpp#L271):

``` .cpp
const xpcc::Vector4f v0;
const xpcc::Matrix4f m0;
const auto v1 = m0 * v0.asTransposedMatrix();
```

fails with the following error:

```
main.cpp: In function ‘int main()’:
[...]

PR #318, opened 2017-12-14
[math][fix]fixed matrix multiplication with non square matrices, fixe…

Maju-Ketchup:
…d LU-Decomposition + added unittests + added simple functions

Was working on some other stuff that does not work-> fixed it and wrote Unittests for it.

Took some time to debug this, most was a simple switch of Height and Width which someone implemented the wrong way around.
I added a LU-Decompostition function if someone just want to solve a simple Ax=b equation without getting the lower and upper matrix.
Additional I wrote a function with which argument returns a permutation Matrix P instead of a Vector.

### DIFF
--- a/src/xpcc/math/test/LUDecomposition_test.cpp
+++ b/src/xpcc/math/test/LUDecomposition_test.cpp
@@ -1,0 +1,64 @@
+#include <xpcc/math/matrix.hpp>
+#include <xpcc/math/lu_decomposition.hpp>
+#include "LUDecomposition_test.hpp"
+
+void
+LUDecompositionTest::testLUD()
+{
+	//https://www.wolframalpha.com/input/?i=LU+Decomposition+of+%7B%7B1,2,3%7D,%7B0,1,2%7D,%7B3,4,6%7D%7D
+
+	const float m[] = {
+		1.f, 2.f, 3.f,
+		0.f, 1.f, 2.f,
+		3.f, 4.f, 6.f
+	};
+
+
+	xpcc::Matrix<float, 3, 3> A(m);
+	xpcc::Matrix<float, 3, 3> l;
+	xpcc::Matrix<float, 3, 3> u;
+	TEST_ASSERT_TRUE(xpcc::LUDecomposition::decompose(A, &l, &u));
+	TEST_ASSERT_EQUALS(l[0][0],  1);
+	TEST_ASSERT_EQUALS(l[0][1],  0);
+	TEST_ASSERT_EQUALS(l[0][2],  0);
+	TEST_ASSERT_EQUALS(l[1][0],  0);
+	TEST_ASSERT_EQUALS(l[1][1],  1);
+	TEST_ASSERT_EQUALS(l[1][2],  0);
+	TEST_ASSERT_EQUALS(l[2][0],  3);
+	TEST_ASSERT_EQUALS(l[2][1], -2);
+	TEST_ASSERT_EQUALS(l[2][2],  1);
+
+	TEST_ASSERT_EQUALS(u[0][0], 1);
+	TEST_ASSERT_EQUALS(u[0][1], 2);
+	TEST_ASSERT_EQUALS(u[0][2], 3);
+	TEST_ASSERT_EQUALS(u[1][0], 0);
+	TEST_ASSERT_EQUALS(u[1][1], 1);
+	TEST_ASSERT_EQUALS(u[1][2], 2);
+	TEST_ASSERT_EQUALS(u[2][0], 0);
+	TEST_ASSERT_EQUALS(u[2][1], 0);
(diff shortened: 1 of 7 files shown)


Maju-Ketchup:
BTW.: this should fix
[Issue #85](https://github.com/roboterclubaachen/xpcc/issues/85) which @ekiwi descriped

salkinium (commented):
Nice! But please fix the commit message length.

Maju-Ketchup:
I'll do it when i have time.
How many characters are allowed?

salkinium:
50 chars on the subject, 72 after that. See https://chris.beams.io/posts/git-commit/#seven-rules.

salkinium:
@Maju-Ketchup as you've noticed, we are currently working on the next generation of this framework (over at [modm-io](https://github.com/modm-io)), which will be licensed under MPLv2 (see below for why). Since this is a non-trivial change to xpcc, would you be ok with licensing your current and future contributions as MPLv2 as well?

> 
>

Maju-Ketchup:
For me it's okay but I have to talk with my employer at first, about this matter.

Maju-Ketchup:
My supervisor says the license changes are okay.
And I cut the commit message.

salkinium (dismissed):
Thanks!

src/xpcc/math/test/LUDecomposition_test.cpp
```diff
@@ -0,0 +1,70 @@
...
+	TEST_ASSERT_EQUALS(b[1][0], -7);
+	TEST_ASSERT_EQUALS(b[2][0], 4);
+	b = xpcc::Matrix<float, 3, 1>(n);
+	xpcc::Matrix<float, 3, 3> l1;
+	xpcc::Matrix<float, 3, 3> u1;
+	xpcc::Matrix<float, 3, 3> p;
+	xpcc::LUDecomposition::decompose(A,&l1,&u1,&p);
+	//TEST_ASSERT_TRUE(l == l1);
+	//TEST_ASSERT_TRUE(u == u1);
+	//TEST_ASSERT_TRUE(p == (xpcc::Matrix<float, 3, 3>::identityMatrix()));
```
strongly-typed: What about these tests? Why do they fail when commented in?
Please reason why they fail or fix it.

strongly-typed:
Could you please rebase on the current develop?
Could you [please add some more whitespace](https://github.com/strongly-typed/xpcc/commit/af9d738c17718508e2a6ff5285b99f433e37a316) to match current coding style?

Anything else looks great and I then would be fine with merging. A last test would be running the unit test on a STM target.

Maju-Ketchup:
Done

src/xpcc/math/test/LUDecomposition_test.cpp
Maju-Ketchup: These test should check if the matrix decomposition should with pivot element should work.
I have choosen an example which was working in Wolfrahmalpha without a permutation matrix.
When no row change is applied the permutation Matrix should be identity matrix E, L, U should be the same as in the example before.
The Algorithm implemented in xpcc does a row changes in this example, I don't know why, but results are the same. 
(Proofed in test:
'TEST_ASSERT_TRUE(xpcc::LUDecomposition::solve(A, &b));'
which calls internal the function I wanted to test with this commented tests i forgot to delete.)
Ax=b <=> PAx = Pb  <=> L'U'x = Pb
Ax=b <=> ELUx =Eb <=> LUx = b

strongly-typed:
Manually merged.
Please take care about [tabs vs. spaces](https://github.com/roboterclubaachen/xpcc/blob/develop/doc/coding_convention.md#layout). 
I will run on a discovery board asap.

strongly-typed:
```
Unittests (Dec 19 2017, 16:30:56)
Machine:  83965d307275
User:     root
Os:       Ubuntu 16.04 xenial
Compiler: arm-none-eabi-g++ (GNU Tools for Arm Embedded Processors 7-2017-q4-major) 7.2.1 20170904 (release) [ARM/embedded-7-branch revision 255204]
Local Git User:
Name:
Email:
Last Commit:
SHA:             c4ed672532d587299427f688d58c67a92eaafa77
Abbreviated SHA: c4ed672
Subject:         [math][fix]fixed matrix multiplication + LU-Decomp

Author:
Name:      Marten Junga
Email:     mjunga@amt.rwth-aachen.de
Date:      Thu Dec 14 15:51:18 2017 +0100
Timestamp: 1513263078

Committer:
Name:      Sascha Schade (strongly-typed)
Email:     stronglytyp3d@gmail.com
Date:      Tue Dec 19 12:53:09 2017 +0100
Timestamp: 1513684389

File Status:
Modified:  1
Added:     0
Deleted:   0
Renamed:   0
Copied:    0
Untracked: 0

Passed 4026 tests
OK!
```
Great!

Issue #316, opened 2017-11-21
[ci] Magic packets for TravisCI


strongly-typed:
The [.travis.dep.sh](https://github.com/roboterclubaachen/xpcc/blob/develop/.travis.dep.sh) downloads two magic packets from the Internet (http://box.xpcc.io/avr-gcc.tar.bz2 and http://box.xpcc.io/boost.tar.bz2).

When, how, and who created these magic tar balls. I only know 1/3 of the answer. There should be a reproducible, documented way how to create these. It's fine for me to rely on ARM providing the toolchain tar balls for Cortex-M.

I was able to build the latest 3.6.1 AVR toolchain in a Docker container.

salkinium:
I built them on the Ubuntu server that hosts them, so that TravisCI doesn’t have to build them over and over again. I already forgot how I did that. I’m ready to remove Travis CI and only use Circle CI.

salkinium:
Would also break when I remove that server or misconfigure the subdomain or something, so we should definitely not rely on it.

strongly-typed:
In TravisCI, we use the compiler you provide. Is seems to be a compiler from Atmel.
`gcc version 5.4.0 (AVR_8_bit_GNU_Toolchain_3.4.4_201706301059)`. The `3.4.4` is definitely wrong. This is caused by a bug in Atmel's build script. Hey Atmel, do you have QA?

Why exactly don't we use the `avr-gcc` we could install in TravisCI's Ubuntu? Because Travis is based on Ubuntu 14.04 and the compiler in 14.04 is too old? Travis seems to stick to good old 14.04.

In CircleCI, we use the [AVR GCC from Ubuntu 16.04](https://github.com/strongly-typed/docker-xpcc-build/blob/master/Dockerfile#L24), which is not a compiler from Atmel.

salkinium:
> In CircleCI, we use the AVR GCC from Ubuntu 16.04, which is not a compiler from Atmel.

My hope was that by using the canonical compiler from Atmel we could avoid issues with the different custom compiled avr-gcc packages by different Linux distributions. However, that requires the end user to actually install the (outdated) Atmel avr-gcc. To my limited experience with our linux devs, that's not what happens, they just install whatever their distributions package manager installs, and so we should also just use the latest avr-gcc from upstream.

The question is really about device coverage, because the upstream avr-gcc fails to compile/assemble for more AVR devices than the Atmel avr-gcc (I assume they don't upstream their changes). 
[...]

PR #322, opened 2017-12-22
[boost] Use C++11 stdlib classes instead of boost.

salkinium:
Trying to reduce the dependence on boost for xpcc for https://github.com/roboterclubaachen/xpcc/issues/316#issuecomment-352525564.

However, we use Boost.Asio for `xpcc::hosted::SerialPort` and there is no replacement in the stdlib for that. So I don't think we can get rid of all of boost.

In addition there is no replacement for `boost::this_thread::disable_interruption`, which breaks the API at the moment.

cc @strongly-typed @dergraaf


emgre:
If you don't want to depend on boost for serial stuff, you can use the standalone Asio library : https://think-async.com/Asio/AsioStandalone

salkinium:
> In addition there is no replacement for `boost::this_thread::disable_interruption`, which breaks the API at the moment.

Never mind, `std::thread` is not interruptible anyways, so I just made this an empty class.

> If you don't want to depend on boost for serial stuff, you can use the standalone Asio library

Great! That will remove our need for all of boost for hosted. I'll refactor it later.

salkinium:
cc @dergraaf Please review these changes, since you wrote most of the affected code. The stdlib changes are based on boost anyways, so this is basically just a fancy find/replace, but please confirm that I implemented `xpcc::rtos::Thread::Lock` correctly!

dergraaf (approved):

PR #328, opened 2018-01-12
[driver] Block device

rleh:
This PR adds a block device interface to xpcc, inspired by mbeds [block device](https://docs.mbed.com/docs/mbed-os-api-reference/en/latest/APIs/storage/block_device/) interface.

Drivers to store data in a SPI SST flash `xpcc::BdSpiFlash`, to store data in a file (on hosted) `xpcc::BdFile` and to store data in RAM (for testing purposes) `xpcc::BdHeap` are included.
Additionally `xpcc::BdMirror` is a virtual block device that behaves like [RAID1](https://en.wikipedia.org/wiki/Standard_RAID_levels#RAID_1) devices.

For every driver a test is included.

The flash chip driver has been tested on a [Nucleo-F429ZI board](https://os.mbed.com/platforms/ST-Nucleo-F429ZI/).

### DIFF
--- a/examples/linux/block_device/mirror/main.cpp
+++ b/examples/linux/block_device/mirror/main.cpp
@@ -1,0 +1,90 @@
+#include <xpcc/architecture.hpp>
+#include <xpcc/debug/logger.hpp>
+
+#include <xpcc/driver/storage/block_device_file.hpp>
+#include <xpcc/driver/storage/block_device_mirror.hpp>
+
+// Set the log level
+#undef	XPCC_LOG_LEVEL
+#define	XPCC_LOG_LEVEL xpcc::log::INFO
+
+void printMemoryContent(const uint8_t* address, std::size_t size) {
+	for (std::size_t i = 0; i < size; i++) {
+		XPCC_LOG_INFO.printf("%x", address[i]);
+	}
+}
+// `testA.bin~` and `testB.bin~` files are created from `SConstruct`
+struct FilenameA {
+	static constexpr const char* name = "testA.bin~";
+};
+
+struct FilenameB {
+	static constexpr const char* name = "testB.bin~";
+};
+
+int
+main()
+{
+	/**
+	 * This example/test writes alternating patterns into
+	 * two `xpcc::BdFile` block device with a size of 1M
+	 * to test the virtual `xpcc::BdMirror` block device.
+	 * The memory content is afterwards read and compared
+	 * to the pattern.
+	 * Write and read operations are done on 64 byte blocks.
+	 */
+
+	constexpr uint32_t BlockSize = 256;
+	constexpr uint32_t MemorySize = 16*1024*1024;
+
+	uint8_t bufferA[BlockSize];
+	uint8_t bufferB[BlockSize];
+	uint8_t bufferC[BlockSize];
+	std::memset(bufferA, 0xAA, BlockSize);
+	std::memset(bufferB, 0x55, BlockSize);
+
+	xpcc::BdMirror<xpcc::BdFile<FilenameA, MemorySize>, xpcc::BdFile<FilenameB, MemorySize>> storageDevice;
+
+	if(!RF_CALL_BLOCKING(storageDevice.initialize())) {
+		XPCC_LOG_INFO << "Error: Unable to initialize device.";
+		exit(1);
+	}
+
+	XPCC_LOG_INFO << "Starting memory test!" << xpcc::endl;
+
+	for(uint16_t iteration = 0; iteration < 10; iteration++) {
+		uint8_t* pattern = (iteration % 2 == 0) ? bufferA : bufferB;
+
+		if(!RF_CALL_BLOCKING(storageDevice.erase(0, MemorySize))) {
+			XPCC_LOG_INFO << "Error: Unable to erase device.";
+			exit(1);
+		}
+
+		for(uint32_t i = 0; i < MemorySize; i += BlockSize) {
+			if(!RF_CALL_BLOCKING(storageDevice.program(pattern, i, BlockSize))) {
+				XPCC_LOG_INFO << "Error: Unable to write data.";
+				exit(1);
+			}
+		}
+
+		for(uint32_t i = 0; i < MemorySize; i += BlockSize) {
+			if(!RF_CALL_BLOCKING(storageDevice.read(bufferC, i, BlockSize))) {
+				XPCC_LOG_INFO << "Error: Unable to read data.";
+				exit(1);
+			}
+			else if(std::memcmp(pattern, bufferC, BlockSize)) {
+				XPCC_LOG_INFO << "Error: Read '";
+				printMemoryContent(bufferC, BlockSize);
+				XPCC_LOG_INFO << "', expected: '";
+				printMemoryContent(pattern, BlockSize);
+				XPCC_LOG_INFO << "'." << xpcc::endl;
+				break;
+			}
+		}
+		XPCC_LOG_INFO << ".";
+	}
+
+	XPCC_LOG_INFO << xpcc::endl << "Finished!" << xpcc::endl;
+
+	return 0;
+}
--- a/examples/linux/block_device/ram/main.cpp
+++ b/examples/linux/block_device/ram/main.cpp
@@ -1,0 +1,75 @@
+#include <xpcc/architecture.hpp>
+#include <xpcc/debug/logger.hpp>
+
+#include <xpcc/driver/storage/block_device_heap.hpp>
+
+// Set the log level
+#undef	XPCC_LOG_LEVEL
+#define	XPCC_LOG_LEVEL xpcc::log::INFO
+
+void printMemoryContent(const uint8_t* address, std::size_t size) {
+	for (std::size_t i = 0; i < size; i++) {
+		XPCC_LOG_INFO.printf("%x", address[i]);
+	}
+}
+
+int
+main()
+{
+	/**
+	 * This example/test writes alternating patterns into
+	 * a `xpcc::BdHeap` block device with a size of 1M.
+	 * The memory content is afterwards read and compared
+	 * to the pattern.
+	 * Write and read operations are done on 64 byte blocks.
+	 */
+
+	constexpr uint32_t BlockSize = 64;
+	constexpr uint32_t MemorySize = 1024*1024;
+
+	uint8_t bufferA[BlockSize];
+	uint8_t bufferB[BlockSize];
+	uint8_t bufferC[BlockSize];
+	std::memset(bufferA, 0xAA, BlockSize);
+	std::memset(bufferB, 0x55, BlockSize);
+
+	xpcc::BdHeap<MemorySize> storageDevice;
+	if(!RF_CALL_BLOCKING(storageDevice.initialize())) {
+		XPCC_LOG_INFO << "Error: Unable to initialize device.";
+		exit(1);
+	}
+	RF_CALL_BLOCKING(storageDevice.erase(0, MemorySize));
+
+	XPCC_LOG_INFO << "Starting memory test!" << xpcc::endl;
+
+	for(uint16_t iteration = 0; iteration < 1000; iteration++) {
+		uint8_t* pattern = (iteration % 2 == 0) ? bufferA : bufferB;
+
+		for(uint32_t i = 0; i < MemorySize; i += BlockSize) {
+			if(!RF_CALL_BLOCKING(storageDevice.write(pattern, i, BlockSize))) {
+				XPCC_LOG_INFO << "Error: Unable to write data.";
+				exit(1);
+			}
+		}
+
+		for(uint32_t i = 0; i < MemorySize; i += BlockSize) {
+			if(!RF_CALL_BLOCKING(storageDevice.read(bufferC, i, BlockSize))) {
+				XPCC_LOG_INFO << "Error: Unable to read data.";
+				exit(1);
+			}
+			else if(std::memcmp(pattern, bufferC, BlockSize)) {
+				XPCC_LOG_INFO << "Error: Read '";
+				printMemoryContent(bufferC, BlockSize);
+				XPCC_LOG_INFO << "', expected: '";
+				printMemoryContent(pattern, BlockSize);
+				XPCC_LOG_INFO << "'." << xpcc::endl;
+				exit(1);
+			}
+		}
+		XPCC_LOG_INFO << ".";
+	}
+
+	XPCC_LOG_INFO << xpcc::endl << "Finished!" << xpcc::endl;
+
+	return 0;
+}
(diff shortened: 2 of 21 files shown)


examples/linux/block_device/mirror/main.cpp
```diff
@@ -0,0 +1,90 @@
...
+#define	XPCC_LOG_LEVEL xpcc::log::INFO
+
+void printMemoryContent(uint8_t* address, std::size_t size) {
+	for (std::size_t i = 0; i < size; i++) {
+		XPCC_LOG_INFO.printf("%x", address[i]);
+	}
+}
+
+struct FilenameA {
+	static constexpr const char* name = "testA.bin";
```
salkinium: Did you mean to commit the `examples/linux/block_device/mirror/testA.bin` and `testB.bin` files below?

src/xpcc/architecture/interface/block_device.hpp
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * Interface of a Block Device
+ *
+ * Access to storage devices like flash chips
+ *
+ * @author	Raphael Lehmann
+ * @ingroup	storage
+ */
+#ifdef __DOXYGEN__
```
salkinium: Shouldn't this be inside the class? And the drivers then inherit from this class.

src/xpcc/architecture/interface/block_device.hpp
```diff
@@ -0,0 +1,104 @@
...
+
+	/** Read data from one or more blocks
+	 *
+	 *  @param buffer	Buffer to read data into
+	 *  @param address	Address to begin reading from
+	 *  @param size		Size to read in bytes (multiple of read block size)
+	 *  @return			True on success
+	 */
+	xpcc::ResumableResult<bool>
+	read(uint8_t *buffer, uint32_t address, size_t size);
```
salkinium: Maybe add a `using address_t = uint32_t;` to use here and then specialize this for implementations.

src/xpcc/driver/storage/block_device_heap.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+#include <xpcc/processing/resumable.hpp>
+
+namespace xpcc
+{
+
+/**
+ * \brief	Block device using heap (RAM)
+ *
+ * Usually not persistant, but at least useful for testing
```
salkinium: This is useful for battery backed SRAM, some stm32 have up to 4kB of this stuff. Very nice!

src/xpcc/driver/storage/block_device_file.hpp
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * The file is part of the xpcc library and is released under the 3-clause BSD
+ * license. See the file `LICENSE` for the full license governing this code.
+ */
+// ----------------------------------------------------------------------------
+
+#ifndef XPCC_BLOCK_DEVICE_FILE_HPP
+#define XPCC_BLOCK_DEVICE_FILE_HPP
+
+#include <xpcc/processing/resumable.hpp>
```
salkinium: This and all other drivers should include `<xpcc/architecture/interface/block_device.hpp>` and inherit from it.

src/xpcc/architecture/interface/block_device.hpp
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * Interface of a Block Device
+ *
+ * Access to storage devices like flash chips
+ *
+ * @author	Raphael Lehmann
+ * @ingroup	storage
+ */
+#ifdef __DOXYGEN__
+class BlockDevice : protected xpcc::NestedResumable<3>
```
salkinium: Don't inherit the interface from `*Resumable`, since that is implementation detail.

src/xpcc/driver/storage/block_device_file.hpp
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * \brief	Block device using a file
+ *
+ * Persistant and useful for testing and hosted code
+ *
+ * \ingroup	driver_storage
+ * \author	Raphael Lehmann
+ */
+template <class Filename, size_t DeviceSize>
+class BdFile : protected xpcc::NestedResumable<3>
```
salkinium: Inherit from `public xpcc::BlockDevice` too.

src/xpcc/driver/storage/block_device_spiflash_impl.hpp
```diff
@@ -0,0 +1,213 @@
...
+ *
+ * The file is part of the xpcc library and is released under the 3-clause BSD
+ * license. See the file `LICENSE` for the full license governing this code.
+ */
+// ----------------------------------------------------------------------------
+
+#ifndef XPCC_BLOCK_DEVICE_SPIFLASH_HPP
+#error	"Don't include this file directly, use 'block_device_spiflash.hpp' instead!"
+#endif
+#include "block_device_spiflash.hpp"
```
salkinium, line 13: nope.

src/xpcc/driver/storage/block_device_mirror_impl.hpp
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * The file is part of the xpcc library and is released under the 3-clause BSD
+ * license. See the file `LICENSE` for the full license governing this code.
+ */
+// ----------------------------------------------------------------------------
+
+#ifndef XPCC_BLOCK_DEVICE_MIRROR_HPP
+	#error	"Don't include this file directly, use 'block_device_mirror.hpp' instead!"
+#endif
+#include "block_device_mirror.hpp"
```
salkinium, line 13: nope.

src/xpcc/driver/storage/block_device_file_impl.hpp
```diff
@@ -0,0 +1,96 @@
...
+ *
+ * The file is part of the xpcc library and is released under the 3-clause BSD
+ * license. See the file `LICENSE` for the full license governing this code.
+ */
+// ----------------------------------------------------------------------------
+
+#ifndef XPCC_BLOCK_DEVICE_FILE_HPP
+	#error	"Don't include this file directly, use 'block_device_file.hpp' instead!"
+#endif
+#include "block_device_file.hpp"
```
salkinium, line 13: nope.

src/xpcc/driver/storage/block_device_heap.hpp
```diff
@@ -0,0 +1,95 @@
...
+	 *  @param address	Address of block to begin erasing
+	 *  @param size		Size to erase in bytes (multiple of read block size)
+	 *  @return			True on success
+	 */
+	xpcc::ResumableResult<bool>
+	erase(uint32_t address, size_t size);
+
+	/** Writes data to one or more blocks after erasing them
+	*
+	*  The blocks are erased proir to being programmed
```
salkinium: pior

src/xpcc/driver/storage/block_device_mirror.hpp
```diff
@@ -0,0 +1,119 @@
...
+	 *  @param address	Address of block to begin erasing
+	 *  @param size		Size to erase in bytes (multiple of read block size)
+	 *  @return			True on success
+	 */
+	xpcc::ResumableResult<bool>
+	erase(uint32_t address, size_t size);
+
+	/** Writes data to one or more blocks after erasing them
+	*
+	*  The blocks are erased proir to being programmed
```
salkinium: prior

salkinium (changes requested):
Very nice interface and implementation. The drivers need to inherit the interface though for doxygen to understand this relationship.

salkinium:
Oh, and Travis is broken, just ignore it.

examples/linux/block_device/mirror/main.cpp
rleh: Yes, because otherwise the `test*.bin` files have to be created manually by the user (or CI) before running the example.
The is no nice way to create the files from C++.

Idea: Create the file in the SConstruct:
```python
xpccpath = '../../../..'
# create empty `test.bin~` file (if it does not exist)
open('test.bin~', 'a').close()
exec(compile(open(xpccpath + '/scons/SConstruct', "rb").read(), xpccpath + '/scons/SConstruct', 'exec'))
```
(Filename changed to `test.bin~` so that it is ignored by [gitignore](https://github.com/roboterclubaachen/xpcc/blob/develop/.gitignore).)

(Edit: filename `test.bin~`)

src/xpcc/driver/storage/block_device_mirror_impl.hpp
rleh: Why? This includes don't hurt and make QtCreators syntax highlighting and auto-completion work.

src/xpcc/driver/storage/block_device_file_impl.hpp
rleh: See below.

src/xpcc/driver/storage/block_device_mirror_impl.hpp
salkinium: Yeah, ok, that's a good enough reason.

salkinium (dismissed):
Nice!

src/xpcc/driver/storage/block_device_spiflash.hpp
```diff
@@ -0,0 +1,230 @@
...
+ * \tparam Cs			The GpioOutput pin connected to the flash chip select
+ * \tparam flashSize	Flash chip size in byte
+ *
+ * The `read()`, `erase()`,`program()` and `write()` methodes wait for
+ * the chip to finish writing to the flash.
+ *
+ * \ingroup	driver_storage
+ * \author	Raphael Lehmann
+ */
+template <typename Spi, typename Cs, size_t flashSize>
```
chris-durand: On AVR size_t is a 16 bit unsigned int. This wouldn't allow you to use a flash larger than 65535 bytes. You could take a plain uint32_t or something like `using flash_size_t = ...` and define it to uint32_t on microcontrollers and size_t on hosted to also support 4GB+ files. The same applies for the size_t function parameters.

examples/linux/block_device/mirror/main.cpp
```diff
@@ -0,0 +1,90 @@
...
+#include <xpcc/debug/logger.hpp>
+
+#include <xpcc/driver/storage/block_device_file.hpp>
+#include <xpcc/driver/storage/block_device_mirror.hpp>
+
+// Set the log level
+#undef	XPCC_LOG_LEVEL
+#define	XPCC_LOG_LEVEL xpcc::log::INFO
+
+void printMemoryContent(uint8_t* address, std::size_t size) {
```
chris-durand: const uint8_t*

examples/linux/block_device/ram/main.cpp
```diff
@@ -0,0 +1,75 @@
+#include <xpcc/architecture.hpp>
+#include <xpcc/debug/logger.hpp>
+
+#include <xpcc/driver/storage/block_device_heap.hpp>
+
+// Set the log level
+#undef	XPCC_LOG_LEVEL
+#define	XPCC_LOG_LEVEL xpcc::log::INFO
+
+void printMemoryContent(uint8_t* address, std::size_t size) {
```
chris-durand: const uint8_t*

src/xpcc/architecture/interface/block_device.hpp
```diff
@@ -0,0 +1,105 @@
...
+	xpcc::ResumableResult<bool>
+	write(const uint8_t *buffer, uint32_t address, size_t size);
+
+public:
+	/** Minimum block sizes for block device operations
+	 *
+	 * Most persistent storage devices require `program()`,  `erase()`
+	 * and sometimes `read()` operations to be executed on fixed sized blocks.
+	 */
+	static constexpr std::size_t BlockSizeRead = 1;
```
chris-durand: Is there a reason for mixing std::size_t and plain size_t? It is guaranteed to be the same type anyway (since C++11).

src/xpcc/driver/storage/block_device_mirror.hpp
```diff
@@ -0,0 +1,121 @@
...
+	static constexpr std::size_t BlockSizeRead = BlockDeviceA::BlockSizeRead;
+	static constexpr std::size_t BlockSizeWrite = std::max(BlockDeviceA::BlockSizeWrite, BlockDeviceB::BlockSizeWrite);
+	static constexpr std::size_t BlockSizeErase = std::max(BlockDeviceA::BlockSizeErase, BlockDeviceB::BlockSizeErase);
+
+public:
+	/** Direct access to the BlockDeviceA
+	*
+	*  @return	BlockDeviceA
+	*/
+	inline BlockDeviceA getBlockDeviceA() {return blockDeviceA;};
```
chris-durand: Returning a copy is probably not what you intended. I suppose you meant BlockDeviceA&. Probably being able to copy an spi flash driver object is a bad idea. Maybe deleting the copy constructor and copy assignment operator would prevent bugs like that.

src/xpcc/driver/storage/block_device_mirror.hpp
```diff
@@ -0,0 +1,121 @@
...
+
+	/** Direct access to the BlockDeviceB
+	*
+	*  @return	BlockDeviceB
+	*/
+	inline BlockDeviceB getBlockDeviceB() {return blockDeviceB;};
```
chris-durand: BlockDeviceB&

chris-durand (changes requested):

src/xpcc/driver/storage/block_device_spiflash.hpp
rleh: > This wouldn't allow you to use a flash larger than 65535 bytes.

No. You are just unable to write, read or delete >64k memory at once. `address` is a `uint32_t`, so you can write/read to/from a <=4GBit flash chip in <=64kBit blocks.

For write and read operations you would need a >64k buffer in your AVRs RAM, but erasing a complete 64MBit flash chip at once on AVR is more realistic. Hmmm...

rleh: Sorry, I missed that this `size_t` is the template parameter and not an read/write/erase function argument.

I'll change the `flashSize` type to `uint32_t` here.

src/xpcc/architecture/interface/block_device.hpp
rleh: No. I'll change all `size_t` to `std::size_t`.

(`std::size_t` has 385 vs `size_t` has only 184 occurrences in xpcc.)

src/xpcc/architecture/interface/block_device.hpp
rleh: Done.

src/xpcc/driver/storage/block_device_heap.hpp
rleh: For this use-case `BdHeap` is now extended to optionally use externally supplied memory instead of allocation memory. [See here.](https://github.com/roboterclubaachen/xpcc/pull/328/commits/92bfb846d58c66df1e13bd8fb9c3270a510a9bb7#diff-a9fb6f431c1838169ef8fe59f70ef89fR42)

salkinium (approved):

PR #329, opened 2018-01-12
Should not assume PORTA is always going to be used

acristoffers:
I wrote code there to enable PORTA because it's the one I'm using. I forgot to take it out of the class.


salkinium:
Thank gosh we _vigorously_ review all contributions, amiright ;-P

PR #331, opened 2018-01-13
Fix comments for enablePll calls.

Sh4rK:


Sh4rK:
Seems like GitHub is still waiting for Travis for some reason.

Edit: I mean I saw that Travis was removed in a recent commit, but something is not working correctly and GitHub is still expecting results from it.

salkinium (approved):
Good catch, lot's of copy-pasting the same mistake there.

salkinium:
> Seems like GitHub is still waiting for Travis for some reason.

I removed the integration in the repo settings, it should dissappear on the next push.

PR #332, opened 2018-01-14
Remove dependency on "future" python package.

Sh4rK:
One less package to worry about is always good :)


salkinium:
cc @mhthies @strongly-typed

mhthies:
Nice idea. I didn't think about that.

salkinium (approved):
I assume this is approved then.

salkinium:
Merged manually.

PR #333, opened 2018-01-14
Fix TempFileMungeWindows Python 3 compatibility.

Sh4rK:
See relevant change in SCons: https://github.com/SCons/scons/commit/f51b065a8e4c009d58038d96b27eb0a6f7fb4189

### DIFF
--- a/scons/site_tools/gcc_windows.py
+++ b/scons/site_tools/gcc_windows.py
@@ -92,8 +92,8 @@ class TempFileMungeWindows(object):
         native_tmp = SCons.Util.get_native_path(os.path.normpath(tmp))
         
         args = list(map(SCons.Subst.quote_spaces, cmd[1:]))
-        output = " ".join(args).replace("\\", "\\\\")
-        os.write(fd, output + "\n")
+        output = " ".join(args).replace("\\", "\\\\") + "\n"
+        os.write(fd, bytearray(output, "utf-8"))
         os.close(fd)
         if SCons.Action.print_actions:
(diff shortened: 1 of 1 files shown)


salkinium:
Could you remove the whitespace changes? They are quite noisy.
We should probably have one commit were we strip all the whitespace, but I'm reluctant, since it would touch so many files and add quite a bit of size to the history. 🙁

Sh4rK:
Done, although the `.editorconfig` file in the root dir clearly has `trim_trailing_whitespace = true` in it :)

Removing all trailing whitespace in one huge commit would be satisfying, but for the reason you mentioned, I actually think that getting rid of them "naturally" (like how this commit originally did) is the better option. I agree that it introduces a bit of noise, but diff tools can be configured to ignore these (for example you can append `?w=1` to the end of a github diff url to ignore whitespace).

salkinium (approved):
I can't test it (without unreasonable effort), so if it work for you that's good enough for me ;-P

salkinium:
> I actually think that getting rid of them "naturally" (like how this commit originally did) is the better option. I agree that it introduces a bit of noise, but diff tools can be configured to ignore these

I'd like to keep the changes in xpcc as clean and minimum as possible so that I have less work applying these changes to modm. I already have so resolve dozens of merge conflicts manually each time, I don't need more ;-P

Sh4rK:
If modm will eventually replace xpcc, then not doing anything about unneeded whitespace is probably the best option then.

Edit: And yes, it works for me :)

salkinium:
Merged manually.

PR #320, opened 2017-12-19
[scons] Enable C++14 for all targets by default.

salkinium:
Fixes #29.

cc @strongly-typed @ekiwi @daniel-k @chrism333 @dergraaf

PR #335, opened 2018-01-17
[driver] Block device: Add missing include

rleh:
In `driver/block_device_mirror.hpp` `std::max()` is used, but `#include <algorithm>` was missing.

Additionally `std::min()` and `std::max()` in the mock-up stdc++ implementation are now *constexpr* and a example for AVR was added (but not tested in hardware).

Furthermore `BlockDevice::DeviceSize` was added.

### DIFF
--- a/examples/avr/block_device_mirror/SConstruct
+++ b/examples/avr/block_device_mirror/SConstruct
@@ -1,0 +1,5 @@
+# path to the xpcc root directory
+xpccpath = '../../..'
+
+# execute the common SConstruct file
+exec(compile(open(xpccpath + '/scons/SConstruct', "rb").read(), xpccpath + '/scons/SConstruct', 'exec'))
--- a/examples/avr/block_device_mirror/main.cpp
+++ b/examples/avr/block_device_mirror/main.cpp
@@ -1,0 +1,96 @@
+#include <xpcc/architecture.hpp>
+
+#include <xpcc/driver/storage/block_device_heap.hpp>
+#include <xpcc/driver/storage/block_device_mirror.hpp>
+
+#include <xpcc/debug/logger.hpp>
+
+using namespace xpcc::atmega;
+typedef xpcc::avr::SystemClock clock;
+
+// Create a new UART object and configure it to a baudrate of 115200
+Uart0 uart;
+xpcc::IODeviceWrapper< Uart0, xpcc::IOBuffer::BlockIfFull > loggerDevice(uart);
+
+// Set all four logger streams to use the UART
+xpcc::log::Logger xpcc::log::debug(loggerDevice);
+xpcc::log::Logger xpcc::log::info(loggerDevice);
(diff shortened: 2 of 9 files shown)


rleh:
How the heck do we still use avr-gcc-4.9.2 for Circle CI? It was released in October 30, 2014. Gee. [😉](https://github.com/roboterclubaachen/xpcc/pull/320#issuecomment-352915894)

salkinium:
> How the heck do we still use avr-gcc-4.9.2 for Circle CI?

Because Ubuntu and *most* other distributions only ship the official toolchain which is stuck at gcc v5.4.0. That's okish, except [it contains a compiler bug](https://stackoverflow.com/questions/45607450/gcc5-nested-variable-template-is-not-a-function-template), which breaks modm CI, so now I'm building a completely new image using archlinux, which has avr-gcc v7.2.0.

rleh:
GCC 5.4 should be fine for this PR.
But the Circle CI container is based on Ubuntu 14.04 which ships avr-gcc v4.9.2.

salkinium:
I've given up on Arch, it didn't have zmqpp headers. I've instead just compiled [avr-gcc v7.2.0 + binutils for Ubuntu 17.10 here](https://github.com/salkinium/docker-avr-gcc-7) to use in our current Ubuntu based CircleCI images.

daniel-k:
> I've given up on Arch, it didn't have zmqpp headers.

There's an AUR package for that: https://aur.archlinux.org/packages/zmqpp/

salkinium:
> There's an AUR package for that

And… there are issues with linking against system libraries. I don't feel like debugging archlinux on a docker image running in CircleCI. I mean there _is_ a limit to my masochist tendencies.

daniel-k:
> And… there are issues with linking against system libraries.

Interesting, it's running smoothly on my system.

> I don't feel like debugging archlinux on a docker image running in CircleCI. I mean there is a limit to my masochist tendencies.

However, this makes sense too 😄

salkinium:
CircleCI is now running avr-gcc 7, but:
```
algorithm_base.hpp:70:2: note: 'constexpr const T& std::max(const T&, const T&) [with T = long unsigned int]' is not usable as a constexpr function because:
  max(const T& a, const T& b)
  ^
```

Happy debugging ^^

rleh:
```
algorithm_base.hpp:76:2: error: body of constexpr function 'constexpr const T& std::max(const T&, const T&) [with T = long unsigned int]' not a return-statement
```
Because: CircleCI still uses `avr-g++ (GCC) 4.9.2` (see [ci/circleci: hosted-unittests](https://circleci.com/gh/roboterclubaachen/xpcc/1825) -> *Check environment*)

~~@salkinium Could you please initiate a new CircleCI build?~~ I rebased this branch on develop, so the updated CircleCI config file from #337 is included.

salkinium (approved):

Issue #323, opened 2018-01-06
hd44780 not working with STM32F030C8T6


acristoffers:
Hi. I tried to use the Hd44780 class with a 16x2 display, single driver, 4 bit databus.
The display is connected as such:

```cpp
using E  = GpioC13;
using Rw = GpioC14;
using Rs = GpioC15;
using D4 = GpioB9;
using D5 = GpioB8;
using D6 = GpioB7;
using D7 = GpioB6;
```

To use the class, I created a bus starting at `GpioB6`, with width=4 and reversed. The Display library commands are being mostly wiped out by the function `GpioPortBase::write(uint16_t data)`.

While debugging, when I get to this function I have the right value in the data variable. However,
after the line

```cpp
data = xpcc::bitReverse(uint16_t(data << StartPinReversed));
```

the data is not on the right place, so it gets wiped out by the mask (`data & portMask`).

The function `xpcc::Hd44780Base<DATA, RW, RS, E>::initialize(LineMode lineMode)` in the file `hd44780_base_impl.hpp` should write 0011 in the bus (B6, B7, B8, B9) when it calls `Bus<DATA, E, DATA::width>::writeHighNibble(Set8BitBus);`. But when `GPIOB->ODR = (GPIOB->ODR & ~portMask) | (data & portMask);` gets called, the output on the bus is 0000;

How I do the setup:

```cpp
#include <xpcc/architecture.hpp>
#include <xpcc/driver/display.hpp>

using namespace xpcc::stm32;

namespace lcd
{
    using E    = GpioC13;
    using Rw   = GpioC14;
    using Rs   = GpioC15;
    using Bl   = GpioB5;
    using Data = GpioPort<GpioB6, 4, xpcc::GpioPort::DataOrder::Reversed>;
}

using LCD = xpcc::Hd44780<lcd::Data, lcd::Rw, lcd::Rs, lcd::E>;

int main()
{
    lcd::Bl::setOutput(true); # BackLight

    auto display = new LCD(16, 2);

    xpcc::delayMilliseconds(100);
    display->initialize();
    xpcc::delayMilliseconds(100);

    display->setCursor(0, 0);

    *display << "Hello";
    display->flush();
 
    while(true){}

    return 0;
}
```

salkinium:
Hm, I've stepped through 
```cpp
GpioPort<GpioB6, 4, xpcc::GpioPort::DataOrder::Reversed>::write(0b0011)
``` 
[...]

PR #336, opened 2018-01-20
Fix timing in HD44780 driver.

Sh4rK:
Hi!

I have a (supposedly) HD44780 compatible 4x160 LCD display. I don't know anything more exact than this about it, the only thing printed on it is "1604A" which seems to be the code for these kind of 4x16 displays. It looks something like in this datasheet: https://www.farnell.com/datasheets/50577.pdf

When I tried to use it with xpcc's HD44780 driver, it was working fine in 8-bit mode if I compiled the project with no optimization, probably because everything is slower, but it was glitching when ran with optimizations enabled, it was probably missing commands.

I tried some ways to make it work based on the above datasheet, and what can be seen here is what I arrived at after some trial and error. This way the display seem to work perfectly in both 4-bit and 8-bit modes. I'll try to explain how I arrived at these values, but the process wasn't too scientific :)

First, the 10ns delays in the 4-bit mode come from the Th and Tah (hold time) values in the timing diagram (both 10ns). For the 8-bit mode, I looked at Tc (cycle time), which is 1.2us, but waiting 200ns didn't work, so I kind of arbitrarily tried 500ns, and it did work, so I just settled on that. If you are now asking "But what about the 1.2us cycle time in 4-bit mode?", I don't know the answer to that, but it works as it is.

### DIFF
--- a/src/xpcc/driver/display/hd44780_base_impl.hpp
+++ b/src/xpcc/driver/display/hd44780_base_impl.hpp
@@ -188,12 +188,14 @@ xpcc::Hd44780Base<DATA, RW, RS, E>::Bus<Data, Enable, 4>::write(uint8_t data)
 	E::set();
 	xpcc::delayMicroseconds(1);
 	E::reset();
+	xpcc::delayNanoseconds(10);
 
 	DATA::write(data);
 
 	E::set();
 	xpcc::delayMicroseconds(1);
 	E::reset();
+	xpcc::delayNanoseconds(10);
 }
 
 template <typename DATA, typename RW, typename RS, typename E>
@@ -216,6 +218,7 @@ xpcc::Hd44780Base<DATA, RW, RS, E>::Bus<Data, Enable, 4>::read()
 	xpcc::delayMicroseconds(1);
 	data = DATA::read();
 	E::reset();
+	xpcc::delayNanoseconds(10);
 
 	data <<= 4;
 
@@ -223,6 +226,7 @@ xpcc::Hd44780Base<DATA, RW, RS, E>::Bus<Data, Enable, 4>::read()
 	xpcc::delayMicroseconds(1);
 	data |= DATA::read();
 	E::reset();
+	xpcc::delayNanoseconds(10);
 
 	return data;
 }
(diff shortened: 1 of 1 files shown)


salkinium:
Huh, interesting. What's the target you're using and what CPU frequency does it run at? The `delayNanoseconds` doesn't get below 20ns or so (for ~180MHz), due to instruction/data fetching and register management.

@acristoffers: This may solve the issues you were seeing with your display running in 4-bit mode. (related to #323).

salkinium:
> If you are now asking "But what about the 1.2us cycle time in 4-bit mode?", I don't know the answer to that, but it works as it is.

Hm, in retrospect it may have been a bit optimistic to try and do ~1MHz "clock speed" across jumper wires ;-) Probably adding a `initialize<baudrate>()` functions alá Spi/I2c/Uart which then slows down the 1µs timing on the Enable pin might be a good future feature.

Sh4rK:
I'm using a Nucleo-F446RE, at 180MHz (if I configured it correctly). See config here: https://github.com/Sh4rK/xpcc/commit/b5344aad0f8e6e4ae96554fc5ccf2961af039628#diff-c8f0df239e3e11bd63c9d37dd220bfaaR78 (I'm planning to open a PR for this too, but I'm not sure it's correct yet, treat it as a WIP.)

salkinium:
> I'm not sure it's correct yet

Remember to enable overdrive (it increases core voltage a little) at that speed. See [this as example](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/architecture/platform/board/stm32f469_discovery/stm32f469_discovery.hpp#L90-L93). We don't have a `ClockControl` API for it yet tho :(

Sh4rK:
> Remember to enable overdrive (it increases core voltage a little) at that speed.

Could that play any part in the LCD not working? I don't know what's that for, could you explain a bit?

salkinium:
> Could that play any part in the LCD not working? I don't know what's that for, could you explain a bit?

I don't think so, we've forgotten to enable this for years in our robots, didn't have any effect that I know of. But that's the official way™ according to the ST datasheet.

Sh4rK:
Okay, I'll enable it.

salkinium (approved):
Thanks!

PR #339, opened 2018-01-28
Fixes for AMSYS5915 driver

chris-durand:
Fixes:
- Prevent int underflow
- Make I2C address adjustable (You can actually get a sensor address reprogramming kit from AMSYS)
- Expose raw measurement value. The specified value range assumed in getPressure() is just specified as a "typical value" in the datasheet and will deviate for different sensors.


src/xpcc/driver/pressure/amsys5915.hpp
```diff
@@ -72,7 +92,7 @@ struct amsys5915
  *
  * The device runs a cyclic program, which will store a corrected pressure value with
  * 12 bit resolution about every 500 μs within the output registers of the internal ASIC.
- *
```
rleh, line 75: 😒

rleh (approved):

salkinium (approved):

PR #154, opened 2016-06-30
[examples] Add examples for STM32F103C8T6 Minimum System Development …

strongly-typed:
…Board.

PR #341, opened 2018-01-28
STM32F030F4P6 Demo Board

rleh:
This PR adds support for the "new" STM32F030F4P6 "Demo Board" available from [Chinese](http://www.hotmcu.com/stm32f030f4p6-minimum-systerm-boardcortexm0-p-208.html) [online](https://www.aliexpress.com/item/STM32F030F4P6-ARM-CORTEX-M0-Core-Board-Minimum-System-Development-Board-Microcontroller-SWD-ISP-Dual-Download/32823139132.html) [shops](http://www.watterott.com/de/STM32F030F4P6-Minimum-Systerm-Board-Cortex-M0) for less than 1.5U$. (Similar to the STM32F103 Blue Pill Board #154)
Features:
* [STM32F030PF4P6](http://www.st.com/en/microcontrollers/stm32f030f4.html) *Mainstream ARM Cortex-M0 Value line MCU with 16 Kbytes Flash, 48 MHz CPU*
* 8MHz crystal
* 3.3V regualtor (AMS1117)
* Micro USB only for power supply
* Reset button
* BOOT0 jumper for UART bootloader (no DFU 😒)
* UART pin header
* SWD pin header
* 15 GPIO pins
* Orange LED

(Only) the blink example has been tested in hardware.

STM32F030F4P6 Demo Board

Clock tree (8MHz and Pll supply core and all peripherals with 48MHz):
STM32F030F4Px clock configuration

Schematic (in case it disappears from the Chinese online shops) for V1.0, my board is V1.1 and misses J2:
image

### DIFF
--- a/examples/stm32f030f4p6_demo_board/blink/SConstruct
+++ b/examples/stm32f030f4p6_demo_board/blink/SConstruct
@@ -1,0 +1,4 @@
+# path to the xpcc root directory
+xpccpath = '../../..'
+# execute the common SConstruct file
+exec(compile(open(xpccpath + '/scons/SConstruct', "rb").read(), xpccpath + '/scons/SConstruct', 'exec'))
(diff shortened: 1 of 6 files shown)


chris-durand (approved):
👍

salkinium (dismissed):
Nice, I have the V1.0 board, and it works too.

PR #342, opened 2018-01-30
[scons] Python 2 compatibility with configparser

### LINKED ISSUE #338: Python compatibility issues on current dev
Since the current dev is 'upgraded' to Python 3, working with Python 2 throws some errors because the 'ConfigParser' class was renamed to 'configparser'.
I have found a small work around to make it backwards compatible while changing one line in one file.

compatibility xpcc/scons/site_tools/configfile.py
old code(line 32):
`import configparser`
new code:
```
try:
       import configparser
except:
       import ConfigParser as configparser 
```
Should I commit this small to the current dev for backwards compatibility or should it stay the way it is so everyone should force update to Python 3?

rleh:
Fixes #338

### DIFF
--- a/scons/site_tools/configfile.py
+++ b/scons/site_tools/configfile.py
@@ -29,7 +29,10 @@
 import re, os, sys
 import xml.etree.ElementTree as et
 import xml.parsers.expat
-import configparser
+try:
+	import configparser
+except:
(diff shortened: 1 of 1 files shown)


salkinium (approved):
Thanks

PR #344, opened 2018-02-01
[scons] CMake wrapper generator

rleh:
This PR adds basic support for generating CMake projects files.
[Clion](https://www.jetbrains.com/clion/) only supports CMake and reads all project settings from `CMakeLists.txt`.

`scons cmakewrapper` generates a `CMakeLists.txt` with the necessary include paths, compiler options and custom build targets which invoke scons.

#### Todos:
- [x] Compiler flags:
```
set(COMMON_FLAGS "{{ ccflags }}")
```
with `ccflags = ' '.join(env['CCFLAGS'])` evaluates to
```
set(COMMON_FLAGS "-mcpu=$ARM_ARCH $FPU $THUMB -mthumb-interwork $OPTIMIZATION ...")
```
`$ARM_ARCH` etc. should be replaced by scons. (Else: `arm-none-eabi-g++: error: unrecognized argument in option '-mcpu=RM_ARCH'`)
Any idea how to get the CCFLAGS with proper replaced variables from scons?

- [x] ~Currently CMake targets are only generated for `scons` and `scons program`.~
Is there a way to get a list of all available scons targets to generate CMake targets for all scons targets?
Fix: CMake targets for all scons targets are generated from a hand-written list.


salkinium:
This? http://scons.org/doc/0.97/HTML/scons-user/x1161.html

salkinium:
Could you add a CircleCI job that compiles via CMake? Otherwise this won't be maintainable. What additional tools do I need to intall for this? CMake obv, but anything else?

rleh:
CMake is the only additional dependency.
Installing CMake in the Circle CI container is only necessary if the CI should build a project using the CMake wrapper.
We can't do that for `scons qtcreator`, but i will add a call to `scons qtcreator` to CI too.

rleh:
@salkinium Could you please restart the hosted-unittests build once the docker container is ready?

---

~I just tested `scons cmakewrapper` in an external project and got the following error:~
```bash
user@host:/project/dir/src $ scons cmakewrapper
[...]
scons: Building targets ...
scons: *** Do not know how to make File target `cmakewrapper' (/home/rleh/src/rexus-viper-elektronik/software/onboard/src/cmakewrapper).  Stop.
scons: building terminated because of errors.
```
~Any ideas why scons does not recognize the cmakewrapper target here?~

All custom `SConscript` files have to be adapted...

salkinium (approved):
Nice!

Issue #345, opened 2018-02-12
scons fails to build for stm32f0xx devices.


lmoesch:
```
[xpcc\build\stm32f0_discovery\blink\main.o] UnicodeDecodeError : 'charmap' codec can't decode byte 0x81 in position 42: character maps to <undefined>
scons: building terminated because of errors.
```

Get this error for the stm32f0_discovery/blink example, aswell as my regular project code, which builds flawlessly for e.g. the stm32f746g_discovery or the f4 nucleo boards.

salkinium:
Position 42 in `stm32f0_discovery/blink/main.cpp` should be a newline value. Maybe your editor converted something? Hm, but it's weird that it's only for F0 projects.

```cpp
#include␣<xpcc/architecture/platform.hpp>¬
¬
using␣namespace␣Board;¬
¬
int¬
main()¬
{¬
>·······Board::initialize();¬
¬
>·······LedGreen::set();¬
>·······LedBlue::reset();¬
¬
>·······while␣(true)␣{¬
>·······>·······LedGreen::toggle();¬
>·······>·······LedBlue::toggle();¬
>·······>·······xpcc::delayMilliseconds(Board::Button::read()␣?␣500␣:␣1000);¬
>·······}¬
}¬
```

lmoesch:
Jeah, despite the error message, it does not seem to have anything todo with the main.cpp. When I change the build target, the error disappears, so I guess it has some problems with parsing/loading the device files.

I even tried the following code:

`int
main(){
while(1)
{}};`

Still the same error.

Can you reproduce the error on Linux?

lmoesch:
The root of the problem were carelessly used "io.open" calls in the template.py, wich leads to encoding errors, when pythons default encoding ist not utf-8 (which it is unfortunately not under Windows /sigh). I temporarily fixed it, by specifying the encoding directly in all io.open calls. I guess the "correct" way would be trying to get the encoding from file or atleast try-catching all io.open/read calls so that useful error messages can be displayed.

salkinium:
Hm, this sucks, this was probably a side-effect of switching to Python 3, which IIRC has different behavior on file encodings.

ekiwi:
Can we just always use utf-8?
@lmoesch : how did you specify the encoding for the `io.open` call? Should we do this for all code that reads a file to a string?

lmoesch:
[...]

PR #349, opened 2018-04-10
fixed encoding errors in template.py

lmoesch:
As mentioned in #345 , I set encoding to utf-8 to io.open calls in template.py, so that xpcc compiles on Windows machines. 

I tested it on stm32f0/4/7 boards so far.

### DIFF
--- a/scons/site_tools/template.py
+++ b/scons/site_tools/template.py
@@ -1,11 +1,11 @@
 #!/usr/bin/env python3
-# 
+#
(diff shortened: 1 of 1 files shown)


salkinium (dismissed):
Thanks!

salkinium:
What happened here? Why are all the commits from #351 here?

lmoesch:
sorry, I pushed to the wrong branch! Removed the latest commits so it should be fine now :)

salkinium (approved):
Thanks!

PR #350, opened 2018-04-12
[STM32] I2C driver

rleh:
There are two different I2C IPs in STM32 devices, the older is named `i2c1_v1_5` and the newer `i2c2_v1_1` (names from STs CubeMX data).

IP `i2c2_v1_1` is included in the following series:
* STM32F0
* STM32F3
* STM32F4
* STM32F7
* STM32H7
* STM32L0
* STM32L4

IP `i2c1_v1_5` is included in the following series:
* STM32F1
* STM32F2
* STM32F4
* STM32L1

Except for STM32F4 the set are disjoint.
Most STM32F4 devices include the old IP, only some STM32F4 devices (STM32F410, STM32F412, STM32F413, STM32F423, STM32F446) include the new IP.

ToDos:
- [ ] Test in hardware on as much devices as possible
- [ ] Combined error and event interrupts (only STM32F0 and STM32L0 devices)


src/xpcc/architecture/platform/driver/i2c/stm32_i2c2_v1_1/driver.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version='1.0' encoding='UTF-8' ?>
+<!DOCTYPE rca SYSTEM "../../xml/driver.dtd">
+<rca version="1.0">
+	<driver type="i2c" name="stm32_i2c2_v1_1">
+		<!-- Template: file that needs to be turned into a source file -->
+		<template device-family="f0|f7|h7|l0|l4" instances="1,2,3" out="i2c_master_{{id}}.hpp">i2c2_v1_1_master.hpp.in</template>
```
rleh: What is `instances="1,2,3" ` in the `driver.xml` file used for?

chris-durand: The attribute specifies for which instances of a peripheral a particular source file is generated. For example, in the STM32 timer driver this feature is used to select the matching implementation for different types of timers. In most cases all possible ids should be listed. The current value "1,2,3" should probably be "1,2,3,4" to support STM32s with 4 I2Cs.

rleh: So this information is redundant for all drivers except for the stm32 timer?!
The device xml files also specify the *instances* attribute: https://github.com/roboterclubaachen/xpcc/blob/c442b5ce1fc4630a37f2b3bf7e126ead8091f816/src/xpcc/architecture/platform/devices/stm32/stm32f405_407_415_417-i_o_r_v_z-e_g.xml#L138

salkinium:
Hm, what project do you need this for? Could you be convinced to implement this in modm instead? How time-critical is this feature?

I’ve [encoded the peripheral type and other things in modm-devices already](https://github.com/modm-io/modm-devices/blob/develop/tools/generator/dfg/stm32/stm_peripherals.py) and modm [has a native way to access this information in its modules](https://github.com/modm-io/modm/blob/develop/src/modm/platform/i2c/stm32/module.lb).

These kind of non-code changes are some work to port (I haven’t fully ported the L4 I2C driver either yet), so I’d prefer switching over non-RCA project to modm sooner rather than later.

salkinium:
Another thing, the CubeMX version numbers don’t necessarily map to the actual feature set. I think they encode the hardware version, not the “register API” version. That’s why I don’t even use this information in modm-devices (except as a fallback solution), since there are multiple version that map to the same memory map/feature set.

src/xpcc/architecture/platform/driver/i2c/stm32_i2c2_v1_1/driver.xml
salkinium: The device file specifies what instances actually exist, the driver file specifies a filter for these instances. Compare this to [the timer instance filter in modm](https://github.com/modm-io/modm/blob/develop/src/modm/platform/timer/stm32/module.lb#L60-L64).

rleh:
> Hm, what project do you need this for? [...] How time-critical is this feature?

Just a private mini-project with a STM32F030F4P6 board, not time-critical. Software-I2C might be an option, too...
At RCA we have some project that would also profit from Hardware-I2C on STM32F0, the idea of implementing it is not new.

> Could you be convinced to implement this in modm instead?

Likely, but: modm is missing any kind of documentation :unamused: 

> These kind of non-code changes are some work to port (I haven’t fully ported the L4 I2C driver either yet), so I’d prefer switching over non-RCA project to modm sooner rather than later.

Therefore modm has to be somehow usable, documentation or at least a getting-started instead of dead links in the Readme would be nice.

> Another thing, the CubeMX version numbers don’t necessarily map to the actual feature set. I think they encode the hardware version, not the “register API” version. That’s why I don’t even use this information in modm-devices (except as a fallback solution), since there are multiple version that map to the same memory map/feature set.

I just took the names from CubeMx because `stm32l4` is a bad name for a I2C IP included in half of all STM32 devices. I compared the "register API" of the devices using the STM32 CMSIS headers.
Where does modm get its information from?

salkinium:
> Therefore modm has to be somehow usable, documentation or at least a getting-started instead of dead links in the Readme would be nice.

I thought [I'd get away with it](https://github.com/modm-io/modm/issues/3) 🙃

> I just took the names from CubeMx because stm32l4 is a bad name for a I2C IP included in half of all STM32 devices. I compared the "register API" of the devices using the STM32 CMSIS headers.
Where does modm get its information from?

I manually compared the memory maps in the Reference Manuals. I looked into automating this process, however, the three primary sources of memory map data (SVD file, CMSIS header and RM) are not fully identical for all devices. Sometimes they miss information that others have, or have conflicting information or are wildy differently formatted. So I gave up and used the RM as the definitive reference point. But basically the same principle.

salkinium:
While I was looking at the RMs anyways, I also grabbed information about feature set differences of peripherals. For example the [F0, F3 and F7 can do 1Mbps I2C Fast-Mode, but the others cannot](https://github.com/modm-io/modm-devices/blob/develop/tools/generator/dfg/stm32/stm_peripherals.py#L298-L325). This isn't visible from the memory map alone.

Similarly the `features` data also encodes features that are added incrementally to new peripherals without a completely new memory map. For example, there is one common SPI peripheral implementation, but in some newer devices, it has more features, like [variable transfer size](https://github.com/modm-io/modm-devices/blob/develop/tools/generator/dfg/stm32/stm_peripherals.py#L141), which is then [queried directly in the driver template](https://github.com/modm-io/modm/blob/develop/src/modm/platform/spi/stm32/spi_base.hpp.in#L102).

I feel like this semantical encoding of a feature set is easier to use than just a version number, which do not even adhere to the semantic versioning rules when mapped to the changes in the memory map. So you'll need another table to map the version to the actual feature set, at which point you can just skip the version directly.

salkinium:
What the status of this? Do you want to implement this in modm instead? The modm toolbox is much more powerful than this one.

rleh:
I will implement and test this in modm.

PR #351, opened 2018-04-13
Fixed shared UARTirq for stm32f030cc

### LINKED ISSUE #88: STM32F0(72): Missing interrupt handler for USART 3 and 4
The STM32F072 has a combined interrupt handler for USART 3 and 4 (`USART3_4_IRQHandler()`), but xpcc generates an individual handler for both peripherals.

### LINKED ISSUE #348: Building generated platform UART5/6 fails for STM32F030CC
While including <xpcc/architecture.hpp> the building process fails for STM32F030CC with the following error:

> [...]\build\libxpcc\generated_platform\driver\uart\stm32\uart_hal_5_impl.hpp: In static member function 'static void xpcc::stm32::UsartHal5::enableInterruptVector(bool, uint32_t)':
[...]\build\libxpcc\generated_platform\driver\uart\stm32\uart_hal_5_impl.hpp:146:20: error: 'USART5_IRQn' was not declared in this scope
   NVIC_SetPriority(USART5_IRQn, priority);

This error happens identically for 'USART6_IRQn'

This only happens for the 'CC' variant with more Uarts. E.g. 'C8' builds fine.

lmoesch:
Okay, this is my attempt to somehow fix the issue with shared IRQs on stm32 devices mentioned in #88 and #348 . The fix has two stages.

First new paramters are added to the `driver.xml`. A parameter called `shared_irq` containing the name of the shared UART (As far as I have seen it, this is either USART3_6 or USART3_8) and a parameter called `shared_irq_ids` containing a whitespace seperated list of the UART ids that are shared.

Than these new parameters are used to generate the appropriate code in the `uart.cpp`, `uart.hpp` and `uart_hal_impl.hpp`.

If an UART is shared, than in the `uart.cpp` there is a class member function `::irq()` defined instead of the interrupt ISR. This function is than called by the shared irq to ensure proper functionality. So where is the shared irq defined?

My first attempt was to generate a seperate file cpp/hpp pair which contains the shared irq routine. Unfortunately irqs - although **_KNOWN_** by the arm compiler - are treated as weak functions and are therefore not linked by the linker, and the common workaround is to add in another function into you .cpp and call it somewhere....great!

Long story short - I put it in the uart3.cpp.

For now it only fixes the issue on the stm32f030cc but more stm32 devices can be supported by adding more variants to the `driver.xml`  - of course it would be far more sophisticated to get these parameters from the devicefile directly.

Fixes #348 and fixes #88.

### DIFF
--- a/src/xpcc/architecture/platform/driver/uart/stm32/driver.xml
+++ b/src/xpcc/architecture/platform/driver/uart/stm32/driver.xml
@@ -15,5 +15,18 @@
 		<parameter name="flow" type="bool">false</parameter>
 		<parameter name="tx_buffer" type="int" min="1" max="65534">250</parameter>
 		<parameter name="rx_buffer" type="int" min="1" max="65534">16</parameter>
+		<!-- shared Uart interrupt -->
+		<shared_irq device-family="f0" device-name="030" device-size-id="c">USART3_6</shared_irq>
+		<shared_irq device-family="f0" device-name="070|071|072" device-size-id="b">USART3_4</shared_irq>
+		<shared_irq device-family="f0" device-name="078">USART3_4</shared_irq>
+		<shared_irq device-family="f0" device-name="091" device-size-id="c">USART3_8</shared_irq>
+		<shared_irq device-family="f0" device-name="098">USART3_8</shared_irq>
+		<shared_irq device-family="l0">USART4_5</shared_irq>
+		<shared_irq_ids device-family="f0" device-name="030" device-size-id="c">3 4 5 6</shared_irq_ids>
+		<shared_irq_ids device-family="f0" device-name="070|071|072" device-size-id="b">3 4</shared_irq_ids>
+		<shared_irq_ids device-family="f0" device-name="078">3 4</shared_irq_ids>
+		<shared_irq_ids device-family="f0" device-name="091" device-size-id="c">3 4 5 6 7 8</shared_irq_ids>
+		<shared_irq_ids device-family="f0" device-name="098" >3 4 5 6 7 8</shared_irq_ids>
+		<shared_irq_ids device-family="l0">4 5</shared_irq_ids>
 	</driver>
 </rca>
--- a/src/xpcc/architecture/platform/driver/uart/stm32/uart.cpp.in
+++ b/src/xpcc/architecture/platform/driver/uart/stm32/uart.cpp.in
@@ -14,10 +14,22 @@
 %#
 %% set name = uart ~ id
 %% set hal = uart ~ "Hal" ~ id
+%#
+%% if shared_irq_ids is defined
+%%  set shared_irq_ids = shared_irq_ids.split(" ")
+%% else
+%%  set shared_irq_ids = []
+%% endif
 
 #include "../../../device.hpp"
 #include "uart_hal_{{ id }}.hpp"
 #include "uart_{{ id }}.hpp"
+
+%% if id == 4
+%% for s in shared_irq_ids
+#include "uart_{{ s }}.hpp"
+%% endfor
+%% endif
 
 %% if parameters.buffered
 #include <xpcc/architecture/driver/atomic.hpp>
@@ -192,7 +204,12 @@ xpcc::stm32::{{ name }}::discardReceiveBuffer()
 
 %% if parameters.buffered
 %% set hal = "xpcc::stm32::" ~ hal
+%% if id | string in shared_irq_ids
+void
+xpcc::stm32::{{ name }}::irq()
+%% else
 XPCC_ISR({{ uart | upper ~ id }})
+%% endif
 {
 	if ({{ hal }}::isReceiveRegisterNotEmpty()) {
 		// TODO: save the errors
@@ -210,5 +227,14 @@ XPCC_ISR({{ uart | upper ~ id }})
 			txBuffer.pop();
 		}
 	}
+}
+%% endif
+
+%% if id == 4
+XPCC_ISR({{ shared_irq }})
+{
+%% for s in shared_irq_ids
+    xpcc::stm32::Usart{{ s }}::irq();
+%% endfor
 }
 %% endif
--- a/src/xpcc/architecture/platform/driver/uart/stm32/uart.hpp.in
+++ b/src/xpcc/architecture/platform/driver/uart/stm32/uart.hpp.in
@@ -12,6 +12,12 @@
 %%	set uart = "Uart"
 %% endif
 %% set hal = uart ~ "Hal" ~ id
+%#
+%% if shared_irq_ids is defined
+%%  set shared_irq_ids = shared_irq_ids.split(" ")
+%% else
+%%  set shared_irq_ids = []
+%% endif
 
 #ifndef XPCC_STM32_UART_{{ id }}_HPP
 #define XPCC_STM32_UART_{{ id }}_HPP
@@ -115,6 +121,11 @@ template< 	class SystemClock, uint32_t baudrate,
 
 	static std::size_t
 	discardReceiveBuffer();
+
+%% if id | string in shared_irq_ids
+	static void
+	irq();
+%% endif
 };
 
 }	// namespace stm32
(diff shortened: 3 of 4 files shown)


rleh:
There are some more shared interrupts on other devices according to the CMSIS header files:

| Device | shared interrupt handler |
|-----------------|--------------------|
| stm32f030xc |   USART3_6_IRQn |
| stm32f070xb |   USART3_4_IRQn |
| stm32f071xb |   USART3_4_IRQn |
| stm32f072xb |   USART3_4_IRQn |
| stm32f078xx |   USART3_4_IRQn |
| stm32f091xc |   USART3_8_IRQn |
| stm32f098xx |   USART3_8_IRQn |
| stm32l071xx |   USART4_5_IRQn |
| stm32l072xx |   USART4_5_IRQn |
| stm32l073xx |   USART4_5_IRQn |
| stm32l081xx |   USART4_5_IRQn |
| stm32l082xx |   USART4_5_IRQn |
| stm32l083xx |   USART4_5_IRQn |

```xml
<!-- shared Uart interrupt -->
<shared_irq device-family="f0" device-name="030" device-size-id="c">USART3_6</shared_irq>
<shared_irq device-family="f0" device-name="070|071|072" device-size-id="b">USART3_4</shared_irq>
<shared_irq device-family="f0" device-name="078">USART3_4</shared_irq>
<shared_irq device-family="f0" device-name="091" device-size-id="c">USART3_8</shared_irq>
<shared_irq device-family="f0" device-name="098">USART3_8</shared_irq>
<shared_irq device-family="l0">USART4_5</shared_irq>
```

lmoesch:
It seems that the stm32f/l0xx variants are the only ones with shared UARTirqs right? The "bigger" ones have enough free interrupt vectors?

rleh:
Searching for shared interrupts in the CMSIS header files (e.g. in `ext/st/`) only shows shared interrupts on stm32[fl]0 devices.

```bash
egrep -R "US?ART[0-9]_[0-9]_IRQ" | grep -v define | sed -r "s/.+\/(stm32[flh][0-9a-frts]{3}.{2}).+:(.*IRQn).*/| \1 | \2 |/g" | sort | uniq
```

lmoesch:
You are quick with regex :) Thanks! Than this should be it. Maybe someone is dedicated to find a better way  to define the shared ISR than in uart3.cpp ...

salkinium:
> My first attempt was to generate a seperate file cpp/hpp pair which contains the shared irq routine.
> Maybe someone is dedicated to find a better way to define the shared ISR than in uart3.cpp ...

TL;DR: Compiler and linker are working as expected, but xpcc isn't linking with the `--whole-archive` flag. Of course
[](https://github.com/modm-io/modm/blob/develop/tools/build_script_generator/scons/resources/SConscript.in#L83-L91)

I've been bugged by this behavior about 3 years ago, but didn't understand what it actually was.
I rediscovered it by accident last year and was able to create a minimal example, file internal issue with the compiler team and finally get to talk with the compiler engineer to understand this behavior.

We first compile xpcc into a static library `libxpcc.a` which is linked with the main program. The linker by default only searches for referenced symbols (to reduce link-time in "large" projects) so all other symbols only get indexed if they are inside an object file that contains a referenced symbol.
This is a problem not only for weak symbols, but also when placing data or functions in special memory regions like `xpcc_section(".hardware_init")`, even if the section is marked with `KEEP(.hardware_init*)`.

If you pass the [`--whole-archive` flag to GNU LD](https://sourceware.org/binutils/docs/ld/Options.html#Options) it will load all symbols regardless of reference.
This doesn't mean all symbols are forcefully included, just that all symbols are *known* to the linker, so `--gc-sections` still works just fine:

> `--whole-archive`
>     For each archive mentioned on the command line after the `--whole-archive` option, include every object file in the archive in the link, rather than searching the archive for the required object files.
> This is normally used to turn an archive file into a shared library, forcing every object to be included in the resulting shared library. This option may be used more than once.

The documentation is maximally unhelpful here, notice how it says "normally used to turn an archive file into a shared library". Wat? But I don't want to turn `libxpcc.a` into a shared library, so I better not use this option? Welcome to the hostile world of embedded software development, where even the linker documentation hates you. I could really use a bigger eyeroll emoji than 🙄.

> Unfortunately irqs - although KNOWN by the arm compiler - are treated as weak functions and are therefore not linked by the linker, and the common workaround is to add in another function into you .cpp and call it somewhere....great!

I think you may be misunderstanding how IRQs are implemented in Cortex-M and xpcc. The compiler knows absolutely nothing about IRQs on ARM, they are not treated specially in any way, you can even exchange them at runtime with any other function if the table is in RAM (with `NVIC_SetVector`).

The interrupt vector table is [generated in `startup.c`](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/architecture/platform/driver/core/cortex/startup.c.in#L83-L101) according to the IRQ positions.
The actual [IRQ function is called `{{name}}_IRQHandler`](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/architecture/platform/driver/core/cortex/startup.c.in#L92) as this is the convention in CMSIS, but this is hidden by the `XPCC_ISR` macro.
The `{{name}}_IRQHandler` is [declared as weak and implemented](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/architecture/platform/driver/core/cortex/startup.c.in#L66) via an alias by [`Undefined_Handler`, which raises a assertion](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/architecture/platform/driver/core/cortex/startup.c.in#L41-L46) if an IRQ vector is called without an appropriate handler.

Note that we don't *have* to use weak functions here at all, it's just super convenient to not have to declare all IRQs that xpcc or the user's program are *not* using, but to the compiler there is absolutely no difference between an IRQ and a normal function.

Btw, with `--whole-archive` the interrupt vector table could finally be put into [it's own file in modm](https://github.com/modm-io/modm/blob/develop/src/modm/platform/core/cortex/vectors.c.in). Before, the linker would just not include it, and there goes your reset vector.

I didn't yet add `--whole-archive` to xpcc, since it's a breaking change for some examples and I didn't want to backport modm to xpcc also. For now just leave this in UART3.

PS: The "ARM compiler" is proprietary (either [completely with ARMCC5](https://developer.arm.com/products/software-development-tools/compilers/arm-compiler/downloads/version-5) or [partly with ARMCC6 based on LLVM](https://developer.arm.com/products/software-development-tools/compilers/arm-compiler)) (both are a pain to use for different reasons), we use the ["GNU Arm Embedded Toolchain"](https://developer.arm.com/open-source/gnu-toolchain/gnu-rm) which is completely upstreamed by ARM ;-)

rleh:
> I put it in the uart3.cpp.

Please put it in `uart4.cpp` instead, because STM32L0 devices only include [UART{1,2,4}](https://github.com/roboterclubaachen/xpcc/blob/develop/src/xpcc/architecture/platform/devices/stm32/stm32l071_072_073_081_082_083-c_k_r_v-8_b_z.xml#L113).
All shared uart interrupts include Uart4.

salkinium (changes requested):
Thanks, code looks good, but please squash your commits into just one.

lmoesch:
@salkinium Thanks for the detailed explanation! Just one thing to mention is what I meant with "known" by the compiler is, that Interrupts have the __irq keyword, which _could_ be used to help them getting linked, but as the manual states:

> [...] Therefore, specifying __irq does not affect the behavior of the compiled output. 

@rleh 
> Please put it in uart4.cpp instead, because STM32L0 devices only include UART{1,2,4}.
All shared uart interrupts include Uart4.

Sure no problem!

@salkinium 
> Thanks, code looks good, but please squash your commits into just one.

Sure!

salkinium:
@lmoesch as you've noticed, we are currently working on the next generation of this framework (over at [modm-io](https://github.com/modm-io)), which will be licensed under MPLv2 (see below for why). Since this is a non-trivial change to xpcc, would you be ok with licensing your current and future contributions as MPLv2 as well?

> 
>

lmoesch:
Sure, no problem to provide it under those licenses!

salkinium (dismissed):

PR #352, opened 2018-04-14
Fix compilation for various STM32L4 targets

chris-durand:
This fixes bugs in the HAL that prevent compilation for many STM32L4 targets:

- On L4 the ADC driver assumes the device header defines ADC123_COMMON, which is not true for STM32L4 with a single ADC peripheral.
- Some L4 devices (471, 475, 476, 485, 486) require the ASCR (analog switch) register to be configured when a pin is used as an ADC input. On all other devices the register is not present. Thus, compilation fails. Furthermore the register is not reset when the pin is reconfigured to a different mode.


salkinium (dismissed):
Thanks!

salkinium:
That force push broke CircleCI for some reason 🙄

PR #353, opened 2018-04-21
[driver] CAN: Bit timing calculation for all clock frequencies

rleh:
The lookup table is replaced by a [constexpr function](https://pastebin.com/UfE9U9LB) written by @chris-durand to calculate the bit timings and prescaler for CAN.

### DIFF
--- a/src/xpcc/architecture/platform/driver/can/generic/can_bit_timings.hpp
+++ b/src/xpcc/architecture/platform/driver/can/generic/can_bit_timings.hpp
@@ -1,5 +1,7 @@
 // coding: utf-8
 /* Copyright (c) 2016, Roboterclub Aachen e.V.
+ * Copyright (c) 2018, Christopher Durand
+ * Copyright (c) 2018, Raphael Lehmann
  * All Rights Reserved.
  *
  * The file is part of the xpcc library and is released under the 3-clause BSD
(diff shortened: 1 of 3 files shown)


ekiwi:
:+1:  for the excellent use of C++14!

If you want to have something to check this against, I once wrote a timing calculator in Python: https://gist.github.com/ekiwi/51573729365405fef89c

salkinium (dismissed):
Very nice!

chris-durand:
I would like to suggest to add the concept of a maximum tolerance to the calculator. A tolerance parameter is already part of the CAN peripheral initialize() function but is silently ignored in the STM32 implementation. I'll prepare a patch tonight to fix this.

chris-durand:
[Done](https://github.com/chris-durand/xpcc/tree/feature/better_can_bit_timings).
@rleh Could you please add these changes to the PR?

src/xpcc/architecture/platform/driver/can/generic/can_bit_timings.hpp
```diff
@@ -96,6 +96,13 @@ class CanBitTiming
 	static constexpr uint8_t getSJW() { return BestConfig.sjw; }
 	static constexpr uint8_t getPrescaler() { return BestConfig.prescaler; }
 
+	template<uint16_t tolerance>
+	static constexpr void assertBitrateInTolerance()
+	{
+		static_assert(tolerance >= static_cast<uint16_t>(BestConfig.minError * 1000.f),
+			"The closest available bitrate exceeds the specified maximum tolerance!");
```
salkinium, line 103: Btw, I know I wrote this, but I always found it confusing. I really with you could somehow format this string, it would make things so much clearer. Any suggestions for something clearer? Or is this ok for you?

chris-durand: It's fine with me.

salkinium (approved):
Awesome!

PR #355, opened 2018-04-23
Added initial support for "Black Pill" board.

genbattle:
[This board](http://wiki.stm32duino.com/index.php?title=Black_Pill) is a variant on the more common "Blue Pill" board that is already supported. The only significant change between the two boards is the pin used for the user LED.

I've added support for the board in the most obvious way: by cloning some of the existing support for the blue pill board and modifying it where appropriate to support the new board. I would like to find a way to refactor the common platform information and examples for both boards; this pull request can either serve as a starting point for this conversation, or it can be accepted as it stands to add support for this board.

### DIFF
--- a/examples/stm32f103c8t6_black_pill/blink/SConstruct
+++ b/examples/stm32f103c8t6_black_pill/blink/SConstruct
@@ -1,0 +1,4 @@
+# path to the xpcc root directory
+xpccpath = '../../..'
+# execute the common SConstruct file
+exec(compile(open(xpccpath + '/scons/SConstruct', "rb").read(), xpccpath + '/scons/SConstruct', 'exec'))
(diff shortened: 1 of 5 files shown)


salkinium:
> The only significant change between the two boards is the pin used for the user LED.

And apparently the correct resistor value on USBDP, so the USB bootloader might actually work!

>  I would like to find a way to refactor the common platform information and examples for both boards; this pull request can either serve as a starting point for this conversation, or it can be accepted as it stands to add support for this board.

Let's leave it like this for xpcc and refactor this instead in modm. I've [already extracted the common logger and `modm_abandon` code](https://github.com/modm-io/modm/blob/develop/src/modm/board/board.cpp.in), as it's far easier to do it this in modm.

salkinium (approved):
Thanks!

salkinium:
Note to self: genbattle accepted dual-licensing BSD/MPLv2 already in #221.

PR #356, opened 2018-04-24
[io] fprint: Print NaN and +-Inf correctly

rleh:
Currently `NaN`, `Inf` and `-Inf` are printed as `0.00000e+00`, which can cause confusion.

### DIFF
--- a/src/xpcc/io/test/io_stream_test.cpp
+++ b/src/xpcc/io/test/io_stream_test.cpp
@@ -13,6 +13,18 @@
 #include <stdio.h>	// snprintf
 #include <string.h>	// memset
 
+#if not defined(XPCC__CPU_AVR)
+#include <limits>
+#else
+#include <math.h>
+namespace std {
+template <typename T> struct numeric_limits {
+	static constexpr T quiet_NaN() { return NAN; }
+	static constexpr T infinity() { return INFINITY; }
+};
+}
+#endif
+
 // ----------------------------------------------------------------------------
 // simple IODevice which stores all data in a memory buffer
 // used for testing the output of an IOStream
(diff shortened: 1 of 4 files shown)


salkinium:
This needs some unittests. Can you check what the relation between the writeFloat and printf implementation is? Are they two different implementations?

rleh:
> This needs some unittests.

Done.

> Can you check what the relation between the writeFloat and printf implementation is? Are they two different implementations?

I didn't see that yesterday. There are two different implementations and now both detect NaN and ±Inf.

rleh:
AVR has no libstdc++ and thus `#include <limits>` fails.

I would like to add a `<limits>` file ~like [this](https://git.uclibc.org/uClibc++/tree/include/limits)~* to our xpcc-libstd++, what do you think?

*`static numeric_type infinity () { return -1; } //How do I properly get this?` :unamused:

src/xpcc/io/test/io_stream_test.cpp
```diff
@@ -300,6 +301,28 @@ IoStreamTest::testFloat4()
 	TEST_ASSERT_EQUALS(device.bytesWritten, 12U);
 }
 
+void
+IoStreamTest::testFloat5()
+{
+	char string[] = "nan";
+
+	(*stream) << std::numeric_limits<float>::quiet_NaN();
```
salkinium, line 320: Try `__builtin_nanf("")`.

src/xpcc/io/test/io_stream_test.cpp
```diff
@@ -300,6 +301,28 @@ IoStreamTest::testFloat4()
...
+	TEST_ASSERT_EQUALS_ARRAY(string, device.buffer, 3);
+	TEST_ASSERT_EQUALS(device.bytesWritten, 3U);
+}
+
+void
+IoStreamTest::testFloat6()
+{
+	char string[] = "inf";
+
+	(*stream) << std::numeric_limits<float>::infinity();
```
salkinium, line 331: `__builtin_huge_valf()`.

salkinium:
> I would like to add a <limits> file like this* to our xpcc-libstd++, what do you think?

I'd be very much in favor of adding an entire libc++ to the AVR build (probably gcc libc++ tho), even if it makes it inefficient to use.
Unfortunately I have no clue what the current state of the art for that is, I'm fairly certain people have done this before (probably Arduino?). I'd have to do some research, cos I don't even think the libc for AVR is complete.

salkinium (approved):

PR #358, opened 2018-05-17
Stricter checking of sign and printf format string.

strongly-typed:
Enabling errors on warnings may break robot code and should be tested against RCA’s code base before merging.

### DIFF
--- a/examples/generic/ros/environment/thread_bme280.cpp
+++ b/examples/generic/ros/environment/thread_bme280.cpp
@@ -95,11 +95,11 @@ Bme280Thread::update()
 			int32_t press = data.getPressure();
 			int32_t hum = data.getHumidity();
 			XPCC_LOG_DEBUG << XPCC_FILE_INFO;
-			XPCC_LOG_DEBUG.printf("BME280: Calibrated temperature in 0.01 degree Celsius is: %d\n", temp  );
+			XPCC_LOG_DEBUG.printf("BME280: Calibrated temperature in 0.01 degree Celsius is: %" PRId32 "\n", temp  );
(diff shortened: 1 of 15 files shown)


salkinium (changes requested):
Thanks for the fixes! 
Can you move the scons tools modifying commits into just one separate commit. This makes it easier for me to test this against RCA code and also to port this to modm.

salkinium:
In general, I don't think using PRI is the correct approach here, this is just fighting the tools.
The reference implementation is our printf implementation, which has a very well defined set of sizes.
We should internally "remap" or "size upgrade" the type identifiers for the hosted implementation of printf, instead of doing this unreadable monkey business.
But this may not prevent gcc from throwing warnings about the inttypes.

strongly-typed:
Thanks Niklas, great to hear from you.

Rearranged the commits.

Yes, this is a bit of fighting tools. `PRId16` is not so common, but it is very strict. Most of the warnings/errors came from compiling for arm.

salkinium:
> Enable strict checkts.

typo!1!!

salkinium (approved):
Thanks. I might cherry-pick the commits without the scons commit first, since this might fail some RCA code.

strongly-typed:
I'm currently checking RCA code. Season is over now.
I checked `big`, `little` and `display` against this branch and did not see any issues.

I'm willing to fix these issues. The point of this commit is to avoid these kind of failures in all kind of new code, including modm.

salkinium:
> I'm currently checking RCA code. Season is over now.
> I checked `big`, `little` and `display` against this branch and did not see any issues.

Ok, let's do it then.

